PR #35, opened 2015-01-12
Popup a diff buffer with C-x v RET

mrBliss:
I wanted to switch from [emacs-git-gutter](https://github.com/syohex/emacs-git-gutter) to diff-hl, but I missed the popup buffer functionality, so I added it :smiley:

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -35,6 +35,7 @@
 ;; `diff-hl-revert-hunk'     C-x v n
 ;; `diff-hl-previous-hunk'   C-x v [
 ;; `diff-hl-next-hunk'       C-x v ]
+;; `diff-hl-popup-hunk'      C-x v RET
 ;;
 ;; The mode takes advantage of `smartrep' if it is installed.
 
@@ -118,6 +119,9 @@
 
 (defvar diff-hl-reference-revision nil
   "Revision to diff against.  nil means the most recent one.")
+
+(defvar diff-hl-popup-buffer "*diff-hl:diff*"
+  "Name of the buffer used to display diffs.")
 
 (defun diff-hl-define-bitmaps ()
   (let* ((scale (if (and (boundp 'text-scale-mode-amount)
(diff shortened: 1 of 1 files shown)


dgutov:
Hmm. Are you sure that you need this feature? `C-x v =`, when `diff-hl-mode` is on, already scrolls to and focuses on the current hunk.

Improving that feature would be a bit easier: for instance, somehow better emphasizing the current hunk in the diff buffer.

mrBliss:
> Improving that feature would be a bit easier: for instance, somehow better emphasizing the current hunk in the diff buffer.

That's partly the issue I'm trying to solve, `C-x v =` gives me too much context, I just want to think about the current hunk/diff.

However, adding `(recenter-top-bottom)` after the while loop in `diff-hl-diff-skip-to` makes `C-x v =` more useable to me.

In what other way do you think the current hunk can be better emphasised in the diff buffer?

dgutov:
I dunno, maybe add or highlight stronger the border around it?

An alternative approach would be to tone down all surrounding text (for instance, remove diff colors from it), but `C-x v =` is also used when I want to see the diff for the whole buffer, and in that case it would be an inconvenience.

katspaugh:
Such a pity this PR wasn't merged. I also miss this feature from `git-gutter`.

dgutov:
There's no merging when the submitted drops out of the discussion.

@katspaugh Why, in your opinion, `C-x v =` is not enough?

katspaugh:
@dgutov I tried `C-x v =` and it does the job quite well. When I said I missed this feature from `git-gutter` I actually meant `git-gutter:revert-hunk`. The equivalent command `diff-hl-revert-hunk` doesn't highlight the current hook as good as `C-x v =` does. And then in `C-x v =` you have to press `C-c C-a` and then confirm the undo, while with the revert command you just confirm.

As I said `C-x v =` does work well for me. Thanks for the great package!

dgutov:
> diff-hl-revert-hunk doesn't highlight the current hunk as good as C-x v = does.

But neither of them highlights the hunk. The difference is that `diff-hl-revert-hunk` shows a hollowed out cursor in the diff buffer (which is a bit harder to notice), because the active cursor is displayed in the prompt. Right?

On the other hand, `diff-hl-revert-hunk` shows a more focused hunk (diff is cut at both sides of the current piece of changes). These cuts emphasize it rather well, IME.

katspaugh:
Maybe I'm using a different version (just installed it from Melpa) or an unsupported color-scheme, but here are the screenshots.

`C-x v =` highlights the hunk with brighter colors:
screen shot 2015-09-29 at 3 27 09 pm

While `diff-hl-revert-hunk` is all the same color and I don't see any cut-offs:
screen shot 2015-09-29 at 3 27 30 pm

dgutov:
Thanks for the screenshots. It's actually the result of (awesome but unrelated) `diff-auto-refine-mode`.

Due to cutting, `diff-hl-revert-hunk` ended up having a different hunk refined (the one above; in your example it has no changes, so it's not visible). Should be fixed now.

katspaugh:
Super, thanks!

PR #36, opened 2015-01-13
Add documentation for diff-hl-dired-status-files

holocronweaver:
The original version check for Emacs 25 fails in Emacs 25, always returning `t` regardless of the emacs-version, making the Emacs 25 workaround fail. My proposed version check should be more robust.

I also added an initial function doc. Feel free to change it - I just wanted there to be SOME kind of doc.

### DIFF
--- a/diff-hl-dired.el
+++ b/diff-hl-dired.el
@@ -108,7 +108,7 @@
                    (setq file (replace-regexp-in-string "\\` " "" file))
                    (let ((type (plist-get
                                 '(edited change added insert removed delete
-                                  unregistered unknown ignored ignored)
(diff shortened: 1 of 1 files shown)


dgutov:
This looks wrong: in Emacs 25, the `dir-status-files` has one fewer argument than before. The current code works right in that regard.

holocronweaver:
My revision does not change the logic of the existing code, it just fixes a non-functioning conditional check for the Emacs version.

holocronweaver:
Since you were right and the current code seems to work fine with the the most recent version of Emacs 25, I will create a separate pull request just for the documentation.

PR #37, opened 2015-01-14
Add documentation for diff-hl-dired-status-files.

holocronweaver:
May not be great, but better than no documentation. =)


dgutov:
Indeed. Thanks!

PR #39, opened 2015-02-25
Move diff-hl-margin-side after diff-hl-margin-mode definition

jamessan:
Using Emacs 24.4.1 (in Debian), I cloned this repo into my ~/.emacs.d and added the following to my init.el:

```elisp
(add-to-list 'load-path (locate-user-emacs-file "diff-hl"))
(require 'diff-hl)
(global-diff-hl-mode)
(unless (window-system)
    (require 'diff-hl-margin)
    (diff-hl-margin-mode))
```
This caused the below error when starting Emacs

> Warning (initialization): An error occurred while loading `/home/mccoyj1/.emacs.d/init.el':
>
> Symbol's value as variable is void: diff-hl-margin-mode

Moving the use of `diff-hl-margin-mode` after its definition fixes the problem for me.

### DIFF
--- a/diff-hl-margin.el
+++ b/diff-hl-margin.el
@@ -44,16 +44,6 @@
   "Highlight buffer changes on margin"
   :group 'diff-hl)
 
-(defcustom diff-hl-margin-side 'left
-  "Which margin to use for indicators."
-  :type '(choice (const left)
-                 (const right))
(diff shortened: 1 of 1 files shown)


dgutov:
I don't see that error, but ok, the change is perfectly reasonable. Thanks!

PR #42, opened 2015-04-08
silence byte-compiler

tarsius:
Declaring `diff-hl-margin-side` silences the byte-compiler, but you might also want to check whether you have to define the options more carefully (possibly using `:set-after`).

Also you might wanna change the newline encoding in `diff-hl.el`.

### DIFF
--- a/diff-hl-margin.el
+++ b/diff-hl-margin.el
@@ -38,6 +38,8 @@
 (require 'diff-hl)
 (require 'diff-hl-dired)
 
+(defvar diff-hl-margin-side)
+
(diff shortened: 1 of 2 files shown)


dgutov:
Thanks.

Is calling `declare-function` inside `eval-when-compile` a common practice?

tarsius:
I don't know, I think it's more often placed outside. I have just recently started doing so (ony when there already was an `eval-when-compile` there that I could use) - so far it hasn't caused any problems for me.

dgutov:
Oh, ok.

I'm not sure how to improve the situation with `:set-after` though, pull request welcome.

PR #47, opened 2015-08-05
Declare variable vc-disable-async-diff

tarsius:
This tells the byte-compiler that it is a dynamically bound variable,
preventing the byte-compiler from thinking it is an unused lexically
bound variable.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -207,6 +207,7 @@
     (t (intern (format "diff-hl-bmp-%s" type)))))
 
 (defvar vc-svn-diff-switches)
+(defvar vc-disable-async-diff)
 
 (defmacro diff-hl-with-diff-switches (body)
   `(let ((vc-git-diff-switches nil)
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -437,13 +437,13 @@ in the source file, or the last line of the hunk above it."
   (interactive)
   (diff-hl-next-hunk t))
 
-(define-prefix-command 'diff-hl-command-map)
-
-(let ((map diff-hl-command-map))
-  (define-key map "n" 'diff-hl-revert-hunk)
-  (define-key map "[" 'diff-hl-previous-hunk)
-  (define-key map "]" 'diff-hl-next-hunk)
-  map)
+(defvar diff-hl-command-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map "n" 'diff-hl-revert-hunk)
+    (define-key map "[" 'diff-hl-previous-hunk)
+    (define-key map "]" 'diff-hl-next-hunk)
+    map))
+(fset 'diff-hl-command-map diff-hl-command-map)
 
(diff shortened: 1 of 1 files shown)


dgutov:
I think it's only saying it's unused because this variable has been removed in Emacs master.

Or do you see this warning even in 24.5?

tarsius:
No, the warning really does means that the variable, which should have had dynamic scope, is used as a variable with lexical scope.

```
$ cat a.el
(defvar a-var 'default)

(defun a-fun ()
  (message "a-var: %s" a-var))
$ cat b.el
;;; -*- lexical-binding: t -*-

(defun b-fun ()
  (let ((a-var 'let-bound))
    (a-fun)))
$ emacs -Q --batch -f batch-byte-compile a.el
$ emacs -Q --batch -f batch-byte-compile b.el

In toplevel form:
b.el:3:1:Warning: Unused lexical variable `a-var'

In end of data:
b.el:6:1:Warning: the function `a-fun' is not known to be defined.
$ emacs -Q --batch -l a.el -l b.el -f b-fun  
a-var: let-bound
$ emacs -Q --batch -l a.elc -l b.elc -f b-fun
a-var: default
```

(defun a-fun ()
  (message "a-var: %s" a-var))
$ cat b.el
;;; -*- lexical-binding: t -*-

(defun b-fun ()
  (let ((a-var 'lexical))
    (a-fun)))

tarsius:
I have rebased and pushed a second commit. Feel free to cherry-pick or to ask me for a second pr.

dgutov:
> No, the warning really does means that the variable, which should have had dynamic scope, is used as a variable with lexical scope.

Yes, that's what it would usually mean. But my point is, you're getting the warning because there's no such variable in Emacs master anymore. And in previous versions, its definition should be pulled in by `(require 'vc)`.

The second change is good. Please make a separate PR for it.

tarsius:
> But my point is, you're getting the warning because there's no such variable in Emacs master anymore.

So do you intend to apply this or a similar change once v25.1 is released? If so I think you might as well do it now. If not then you would either have to stop supporting < v25.1 at that point and remove the use of this variable, or live with the warning being shown to users. I suspect you don't intend to do the former, and I don't think the latter would be a good idea, because it's quite scary (if you know or have a sense for what it means) and users cannot know if this is a serious issue or just because that variable was removed in v25.1. It takes quite some time to figure that out. Because of that and because *we* know it is a harmless warning I think it should be suppressed.

dgutov:
> So do you intend to apply this or a similar change once v25.1 is released?

It would create an impression that a variable such as this still exists. Which could lead to a different kind of confusion.

Since you insist, I've pushed a change that should take care of it for all Emacs versions.

tarsius:
> Since you insist, I've pushed a change that should take care of it for all Emacs versions.

Thanks. And that's better than my suggestion.

Issue #33, opened 2014-12-13
Highlight diffs on-the-fly


PythonNut:
The README indicates that showing accurate diffs even if the buffer has been modified, but not saved is a possible feature. This is an issue to track the discussion and development of this feature, if this is deemed worthwhile to implement.

PythonNut:
How do we intend to fix it? I've never encountered this before. Obviously, we do want it to work in current versions. Do we advise it conditionally on the version?

dgutov:
> Obviously, we do want it to work in current versions

Only if it's not too much trouble. Emacs releases come and go. Not everyone has to have access to the latest goodies. :)

> Do we advise it conditionally on the version?

Probably. First we'll have to fix it in the core, and the advice could be based on that fix.

PythonNut:
For `git`:

```emacs
(defadvice vc-git-working-revision
  (around (file) activate preactivate compile)
  "Git-specific version of `vc-working-revision'."
  (vc-git--rev-parse "HEAD"))
```

I haven't used the other `vc` systems to the point where I can intelligently comprehend their code, and checking them all seems to be a daunting task.

dgutov:
Yes, the above should work. But doing just this will make VC show `Git-ffafafafafaf` in the mode-line where it now shows `Git-master`, which would be a step back in functionality.

Maybe the `working-revision` VC command needs a new optional "literal" argument.

PythonNut:
Hm... I suppose the question is what we mean by _revision_. In my mind, a revision must be immutable. `master` is not a revision, but `ffafafafafaf` is. `master` is a branch.

It's probably a little late to fix this without breaking tons of stuff. 

> Maybe the `working-revision` VC command needs a new optional "literal" argument.

That would require us to update all of the backends, right? That would require quite a few people (unless we know anyone who knows a ton of vc systems).

dgutov:
> I suppose the question is what we mean by revision... It's probably a little late to fix this without breaking tons of stuff.

Yes, now what VC means by "revision" is defined by the places it's used in.

> That would require us to update all of the backends, right? That would require quite a few people (unless we know someone who knows a ton of vc systems).

Not if the argument is optional, then the backends can be updated gradually. And the backends that alreay return a "concrete" revision would be unaffected.

PythonNut:
Handling this is a half step more delicate than I planned because `vc` _also_ caches the revisions in memory.

dgutov:
> vc also caches the revisions as lisp attributes

[...]

PR #50, opened 2015-08-13
[WIP] Allow diffing on-the-fly

PythonNut:
See #33 

This isn't quite ready. It doesn't have the following advice for the two `vc` functions.
```emacs
(with-eval-after-load 'vc
  (defun nadvice/vc-working-revision (file &optional backend concrete)
    (if concrete
      (vc-call-backend backend 'working-revision file t)
      (or (vc-file-getprop file 'vc-working-revision)
        (progn
          (setq backend (or backend (vc-responsible-backend file)))
          (when backend
            (vc-file-setprop file 'vc-working-revision
              (vc-call-backend backend 'working-revision file)))))))
  (advice-add #'vc-working-revision :override #'nadvice/vc-working-revision))

(with-eval-after-load 'vc-git
  (defun nadvice/vc-git-working-revision (old-fun file &optional concrete)
    "Git-specific version of `vc-working-revision'."
    (if concrete
      (vc-git--rev-parse (funcall old-fun file))
      (funcall old-fun file)))

  (advice-add #'vc-git-working-revision :around
    #'nadvice/vc-git-working-revision))
```

Will I

1. Commit some variant of these to the Emacs codebase and disallow `diff-hl-flydiff-mode` on all versions up to but not including the patched one?
2. Commit the code to Emacs, and keep some advice in `diff-hl` that:
  1. Switches on the Emacs version?
  2. Switches on some sort of feature detection?

This has passed my initial testing with the above advice, so things seem to be going well.

### DIFF
--- a/diff-hl-flydiff.el
+++ b/diff-hl-flydiff.el
@@ -66,6 +66,11 @@
     (advice-add 'vc-git-mode-line-string :override
       #'diff-hl-flydiff/vc-git-mode-line-string)))
 
+(defun diff-hl-flydiff/working-revision (file)
+  "Like vc-working-revision, but always up-to-date"
+  (vc-file-setprop file 'vc-working-revision
+    (vc-call-backend (vc-backend file) 'working-revision file)))
+
 (defun diff-hl-flydiff-make-temp-file-name (file rev &optional manual)
   "Return a backup file name for REV or the current version of FILE.
 If MANUAL is non-nil it means that a name for backups created by
@@ -114,13 +119,10 @@ This requires the external program `diff' to be in your `exec-path'."
               (if (file-directory-p "/dev/shm/")
                 "/dev/shm/"
                 temporary-file-directory))
-            (rev (diff-hl-flydiff-create-revision
-                   file
-                   (vc-working-revision file
-                     (vc-responsible-backend file)))))
+            (rev (diff-hl-flydiff-create-revision file
+                   (diff-hl-flydiff/working-revision file))))
       (diff-no-select rev (current-buffer) "-U 0" 'noasync
         (get-buffer-create " *diff-hl-diff*")))))
-
 
 (defun diff-hl-flydiff/update (old-fun &optional auto)
   (unless (and auto
--- a/diff-hl-flydiff.el
+++ b/diff-hl-flydiff.el
@@ -1,4 +1,4 @@
-;; Copyright (C) 2012-2013  Free Software Foundation, Inc.
+;; Copyright (C) 2015 Free Software Foundation, Inc.
 
 ;; Author:   Jonathan Hayase <PythonNut@gmail.com>
 ;; URL:      https://github.com/dgutov/diff-hl
--- a/diff-hl-flydiff.el
+++ b/diff-hl-flydiff.el
@@ -132,14 +132,7 @@ This requires the external program `diff' to be in your `exec-path'."
     (funcall old-fun)))
 
 (defun diff-hl-flydiff/modified-p (state)
-  (or
-    (buffer-modified-p)
-    (eq state 'edited)
-    (and (eq state 'up-to-date)
-      ;; VC state is stale in after-revert-hook.
-      (or revert-buffer-in-progress-p
-        ;; Diffing against an older revision.
-        diff-hl-reference-revision))))
+  (buffer-modified-p))
 
 (defun diff-hl-flydiff/update-modified-tick (&rest args)
   (setq diff-hl-flydiff-modified-tick (buffer-modified-tick)))
@@ -154,7 +147,7 @@ This requires the external program `diff' to be in your `exec-path'."
       (advice-add 'diff-hl-update :around #'diff-hl-flydiff/update)
       (advice-add 'diff-hl-overlay-modified :override #'ignore)
 
-      (advice-add 'diff-hl-modified-p :override
+      (advice-add 'diff-hl-modified-p :before-until
         #'diff-hl-flydiff/modified-p)
       (advice-add 'diff-hl-changes-buffer :override
         #'diff-hl-flydiff-buffer-with-head)
--- a/diff-hl-flydiff.el
+++ b/diff-hl-flydiff.el
@@ -1,0 +1,200 @@
+;; Copyright (C) 2012-2013  Free Software Foundation, Inc.
+
+;; Author:   Jonathan Hayase <PythonNut@gmail.com>
+;; URL:      https://github.com/dgutov/diff-hl
+
+;; This file is part of GNU Emacs.
+
+;; GNU Emacs is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; GNU Emacs is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;;; Code:
+
+(require 'diff-hl)
+
+(defvar diff-hl-flydiff-modified-tick 0)
+(defvar diff-hl-flydiff-timer)
+(make-variable-buffer-local 'diff-hl-flydiff-modified-tick)
+
+;; Polyfill concrete revisions for vc-git-working-revision in Emacs 24.4, 24.5
+(when (version<= emacs-version "24.5")
+  (with-eval-after-load 'vc-git
+    (defun vc-git--symbolic-ref (file)
+      (or
+        (vc-file-getprop file 'vc-git-symbolic-ref)
+        (let* (process-file-side-effects
+                (str (vc-git--run-command-string nil "symbolic-ref" "HEAD")))
+          (vc-file-setprop file 'vc-git-symbolic-ref
+            (if str
+              (if (string-match "^\\(refs/heads/\\)?\\(.+\\)$" str)
+                (match-string 2 str)
+                str))))))
+
+    (defun diff-hl-flydiff/vc-git-working-revision (_file)
+      "Git-specific version of `vc-working-revision'."
+      (let (process-file-side-effects)
+        (vc-git--rev-parse "HEAD")))
+
+    (defun diff-hl-flydiff/vc-git-mode-line-string (file)
+      "Return a string for `vc-mode-line' to put in the mode line for FILE."
+      (let* ((rev (vc-working-revision file))
+              (disp-rev (or (vc-git--symbolic-ref file)
+                          (substring rev 0 7)))
+              (def-ml (vc-default-mode-line-string 'Git file))
+              (help-echo (get-text-property 0 'help-echo def-ml)))
+        (propertize (replace-regexp-in-string (concat rev "\\'") disp-rev def-ml t t)
+          'help-echo (concat help-echo "\nCurrent revision: " rev))))
+
+    (advice-add 'vc-git-working-revision :override
+      #'diff-hl-flydiff/vc-git-working-revision)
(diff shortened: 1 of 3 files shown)


dgutov:
1 or 2, makes no difference to me personally. If you go with 2, `i` seems to be the most feasible.

I would rename `nadvice/vc-working-revision` to something like `diff-hl/vc-working-revision`. No reason to hijack the `nadvice` namespace.

That aside, could you think about how hard would it be to mostly constrain the new code to a new file? It would contain the minor mode, the new functions and, ideally, only thin wrappers around some existing functions.

PythonNut:
> Could you think about how hard would it be to mostly constrain the new code to a new file? It would contain the minor mode, the new functions and, ideally, only thin wrappers around some existing functions.

All the code is hosted in a separate file in my config right, now, so it's all a question of how much modification you want of the main file. I assume you don't want to advice your own functions, so how would I "wrap" a function from the main file?

dgutov:
By adding a new dynamic `xxx-function` variable, for example. If that's feasible.

dgutov:
Although using `nadvice` might not be too bad either. It has "local" advices for purposes such as this.

PythonNut:
> By adding a new dynamic xxx-function variable, for example. If that's feasible.

You'd have to do that for:

* `diff-hl-update`
* `diff-hl-changes`
* `diff-hl-overlay-modified`
* `diff-hl-mode`

I don't know how you'd do that for a minor mode, and making four might be a bit overkill.

> Although using `nadvice` might not be too bad either. It has "local" advices for purposes such as this.

You mean buffer local advice, as described here?

> If `PLACE` is a symbol, its `default-value` will be affected.
> Use `(local 'SYMBOL)` if you want to apply `FUNCTION` to `SYMBOL` buffer-locally.

To be honest, I'm a bit biased towards the `nadvice` strategy, since my code is already written in `nadvice`, so don't listen to me.

What do you think?

dgutov:
> I don't know how you'd do that for a minor mode, and making four might be a bit overkill.

`advice-add` on enable and `advice-remove` on disable might do the trick.

> You mean buffer local advice, as described here?

Yes, although it would require a buffer-local minor mode, and diff-hl-flydiff-mode will probably work best as a global one.

> To be honest, I'm a bit biased towards the nadvice strategy, since my code is already written in nadvice, so don't listen to me.

nadvice seems fine to me (especially if all its uses will be localized to the new file). It can be turned into "proper" dynamic variables later.

Anyway, the first step will be to patch the Emacs core.

PythonNut:
> Anyway, the first step will be to patch the Emacs core.

@dgutov do you know any other version control system where the `concrete` distinction makes sense? I'm going to add the parameter to all of the backends, but for most of them it will do nothing.

dgutov:
> @dgutov do you know any other version control system where the concrete distinction makes sense?

Well, both Bazaar and Mercurial have similar concepts of branches, but the respective implementations of `find-revision` decided to simply return the numeric revisions. Probably because they're short enough, as opposed to Git's long-ish hash.

> I'm going to add the parameter to all of the backends, but for most of them it will do nothing.

Why don't you take this discussion to `emacs-devel` or `bug-gnu-emacs`? Maybe some other developers will have better ideas.

PythonNut:
@dgutov thanks for taking over the patch. I think everything is in order now, except for one thing: I can't seem to get the buffer-local advice to work:

```emacs
(defun pass ()
  (interactive))

(defun hi ()
  (interactive)
  (message "hi!"))

(add-function :override (local 'pass) #'hi)
(pass) ;; does nothing
```

dgutov:
Unfortunately, I have misled you about the purpose of `(local 'pass)` in `add-function`. Turns out, `add-function` is for functions stored in variables, and `(local ...)` makes it modify the buffer-local value of a said variable. For instance, as a result of your code above, `pass` is defined as a variable and `(funcall pass)` outputs the message. Sorry. But that makes sense: Emacs doesn't have buffer-local functions, only buffer-local variables.

You'll need to use `advice-add` instead of `add-function`. If the new minor mode is going to be global, there's no need for `(local ...)` anyway. Otherwise, the advice should check if the minor mode is enabled, and act depending on that.

PythonNut:
I think I'll just do a global minor mode. How do the new changes look?

diff-hl-flydiff.el
```diff
@@ -0,0 +1,203 @@
...
+;;; Code:
+
+(require 'diff-hl)
+
+(defvar diff-hl-flydiff-modified-tick 0)
+(defvar diff-hl-flydiff-timer)
+(make-variable-buffer-local 'diff-hl-flydiff-modified-tick)
+
+;; Polyfill concrete revisions for vc-git-working-revision in Emacs 24.4, 24.5
+(when (version<= emacs-version "24.5")
```
dgutov: Better compare to `25.0`. I suspect many users have `emacs-version` at `24.5.1`.

diff-hl-flydiff.el
```diff
@@ -0,0 +1,203 @@
...
+;; along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; This mode enables diffing on-the-fly (i.e. without saving the buffer first)
+;; Toggle in all buffers with M-x diff-hl-flydiff-mode
+
+;;; Code:
+
+(require 'diff-hl)
```
dgutov: `(require 'advice)`, too, maybe? To check that the current Emacs is new enough.

diff-hl-flydiff.el
```diff
@@ -0,0 +1,203 @@
...
+                          (cl-incf line))
+                        (push (list line len type) res))))))
+              (setq diff-hl-flydiff-modified-tick (buffer-modified-tick))
+              (nreverse res)))
+          ((eq state 'added)
+            `((1 ,(line-number-at-pos (point-max)) insert)))
+          ((eq state 'removed)
+            `((1 ,(line-number-at-pos (point-max)) delete))))))))
+
+(defun diff-hl-flydiff/overlay-modified (&rest args))
```
dgutov: You can just use `#'ignore` instead of this function.

diff-hl-flydiff.el
```diff
@@ -0,0 +1,203 @@
...
+               temporary-file-directory)))
+      (diff-no-select rev (current-buffer) "-U 0" 'noasync
+        (get-buffer-create " *diff-hl-diff*")))))
+
+
+(defun diff-hl-flydiff/update (old-fun &optional auto)
+  (unless (and auto
+            (or
+              (= diff-hl-flydiff-modified-tick (buffer-modified-tick))
+              (file-remote-p default-directory)
```
dgutov: This seems problematic. Since the minor mode is global, will it stop `diff-hl-mode` from working properly in remote buffers altogether?

diff-hl-flydiff.el
```diff
@@ -0,0 +1,203 @@
...
+              (not (buffer-modified-p))))
+    (funcall old-fun)))
+
+(defun diff-hl-flydiff/changes (&rest args)
```
dgutov: Is there a way you can advice this without copying the whole definition?

Extract a predicate function and a variable, maybe?

diff-hl-flydiff.el
PythonNut: Sure, thanks.

diff-hl-flydiff.el
PythonNut: I was thinking of deferring loading `nadvice` until I needed it, but I just realized that it's pretty much required immediately.

diff-hl-flydiff.el
PythonNut: No, because the method of actually getting the changes is different. The new method is a superset of the old one, though, so you might consider just making it the default for the regular `diff-hl-mode`. Then I can just use a simple predicate list here.

diff-hl-flydiff.el
PythonNut: No, because the regular method of triggering an update is still around, and for those `auto` is `nil`, so the `file-remote-p` predicate is ignored.

diff-hl-flydiff.el
```diff
@@ -0,0 +1,195 @@
...
+      (advice-add 'diff-hl-update :around #'diff-hl-flydiff/update)
+      (advice-add 'diff-hl-changes :override #'diff-hl-flydiff/changes)
+      (advice-add 'diff-hl-overlay-modified :override #'ignored)
+
+      (setq diff-hl-flydiff-timer
+        (run-with-idle-timer 0.3 t #'diff-hl-update t)))
+
+    (advice-remove 'diff-hl-update #'diff-hl-flydiff/update)
+    (advice-remove 'diff-hl-changes #'diff-hl-flydiff/changes)
+    (advice-remove 'diff-hl-overlay-modified #'ignored)
```
dgutov: `#'ignore`, not `#'ignored`.

diff-hl-flydiff.el
dgutov: I see.

diff-hl-flydiff.el
PythonNut: Fixed, but I can't actually change the commit message for a few hours. Anyway, I think this is still blocked by the Emacs bug.

PythonNut:
@dgutov okay, I think that's everything, with the exception of `diff-hl-flydiff/changes`, which I'm not sure what to do with.

diff-hl-flydiff.el
```diff
@@ -0,0 +1,197 @@
+;; Copyright (C) 2012-2013  Free Software Foundation, Inc.
```
dgutov: Look at the years. ;-)

PythonNut: I copied the header off of `diff-hl-amend.el`. :P

diff-hl-flydiff.el
dgutov: Maybe you should extract `diff-hl-modified-p` and `diff-hl-changes-buffer` from `diff-hl-changes` (do that in `diff-hl.el`), and then advise those.

PythonNut: Sure, that's what I was thinking, but I didn't want to step on `diff-hl.el` unless directed.

PythonNut:
@dgutov  Alright, I've deduplicated the code. Do you see anything else that needs fixing?

dgutov:
Looks better, but I think you should advice `diff-hl-modified-p` with `:before-until` (in order not to copy its definition).

dgutov:
Docs: yes. Please mention it in the About section of the README.

PythonNut:
@dgutov alright, I think this is the last call for feedback.

dgutov:
Thank you, merging now. There are a couple of minor issues left, but I'll fix them myself.

Tried it out, and it's pretty sweet. :thumbsup:

dgutov:
> I'll fix them myself

Maybe not all of them. Do you see README.md get one big "modified" highlight for the whole buffer?

dgutov:
How come the `advice-add` lines for `vc-git-` functions are indented this way? That's not how my Emacs tries to indent them.

dgutov:
If I make a change and save the buffer before `diff-hl-flydiff-timer` manages to kick in, it doesn't do anything, and the highlighting is wrong until the next buffer modification.

dgutov:
And, for completeness, `0.3` should be a user option.

PythonNut:
> Do you see README.md get one big "modified" highlight for the whole buffer?

No? But I'm not quite sure what you're talking about.

> If I make a change and save the buffer before `diff-hl-flydiff-timer` manages to kick in, it doesn't do anything, and the highlighting is wrong until the next buffer modification.

Hm... For some reason I must have missed this.

> And, for completeness, `0.3` should be a user option.

Oh,of course.

dgutov:
> No? But I'm not quite sure what you're talking about.

Here's a video of me messing around with it: http://webm.land/w/cDeF/

The reason is probably CRLF line endings, which Emacs converts to CR when opening the buffer.

Are you on Windows?

dgutov:
Anyway, 8d313db fixed this specific case.

PythonNut:
> Are you on Windows?

No. Hm... strange. I'll also want to add `diff-hl-with-diff-switches` for consistency.

dgutov:
> I'll also want to add diff-hl-with-diff-switches for consistency.

Why? It's only used for `vc-diff-switches`.

PR #51, opened 2015-08-18
Define diff-hl-command-map without destroying existing bindings

tarsius:
Keymaps are usually defined using `defvar` which has the advantage that
it only sets a variables value if that is not defined yet.  As a result
re-evaluating (i.e. during development) a buffer that contains such a
definition does not reset the value, which would lose all user
customization.

Unfortunately `define-prefix-command` does not do that.  So use `defvar`
to define the variable and then also store the in the function cell
using `fset`.


dgutov:
Thanks!

PR #53, opened 2015-09-05
[WIP] Allow diffing on-the-fly round 2

PythonNut:
This should address some of the issues.

However, I can't seem to reproduce saving before the idle timer fires. It's still updating for me.

### DIFF
--- a/diff-hl-flydiff.el
+++ b/diff-hl-flydiff.el
@@ -122,8 +122,12 @@ This requires the external program `diff' to be in your `exec-path'."
                temporary-file-directory))
             (rev (diff-hl-flydiff-create-revision file
                    (diff-hl-flydiff/working-revision file))))
-      (diff-no-select rev (current-buffer) "-U 0 --strip-trailing-cr" 'noasync
-        (get-buffer-create " *diff-hl-diff*")))))
+      (diff-hl-with-diff-switches
+        (diff-no-select rev
+          (current-buffer)
+          "-U 0 --strip-trailing-cr"
+          'noasync
+          (get-buffer-create " *diff-hl-diff*"))))))
 
 (defun diff-hl-flydiff/update (old-fun &optional auto)
   (unless (and auto
--- a/diff-hl-flydiff.el
+++ b/diff-hl-flydiff.el
@@ -37,14 +37,14 @@
   (with-eval-after-load 'vc-git
     (defun vc-git--symbolic-ref (file)
       (or
-        (vc-file-getprop file 'vc-git-symbolic-ref)
-        (let* (process-file-side-effects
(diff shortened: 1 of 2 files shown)


PythonNut:
p.s. I have this in my init:

```emacs
(setq
  lisp-indent-offset 2
  lisp-body-indent 2)
```

IMHO it makes the indentation less consistent but easier to parse (at least for me) and more like other languages.

dgutov:
> However, I can't seem to reproduce saving before the idle timer fires.

That's odd. It reproduces every time I try. Maybe I'll have to dig deeper.

> I have this in my init:

Please don't do that. Emacs Lisp already has well-established indentation offsets, and when you contribute to other projects using non-default indentation settings, you risk adding spurious indentation changes. Or if you don't, I'll probably end up doing them at some point, when editing your code.

PythonNut:
> Please don't do that.

That was a setting I added last year when I started Emacs. it [shouldn't happen again](https://github.com/PythonNut/emacs-config/commit/e8ea15d70d5573181bac9df3e6c61fac43f66a01).

dgutov:
Thank you!

Now, could you explain the reason to use `diff-hl-with-diff-switches` here?

PythonNut:
@dgutov because of [this commit](https://github.com/PythonNut/diff-hl/commit/cbba426c4ac8dff0a8c1b866f1d4f09620cef3c9).

PythonNut:
For some reason canceling the idle timer doesn't work for me. Do you see this?

dgutov:
> @dgutov because of this commit.

I see. Well... it doesn't seem appropriate. AFAICS, we need `diff-switches` and `vc-` bindings in different situations, so there's no need in putting them together.

dgutov:
> For some reason canceling the idle timer doesn't work for me.

Not sure. How do I reproduce this?

PythonNut:
Try turning `diff-hl-flyspell-mode` off. The timer hangs around and starts causing errors because the advice is no longer active.

dgutov:
It gets disabled: I see no errors after turning `diff-hl-flydiff-mode` off, and `timer-idle-list` doesn't list `diff-hl-update` after that either. Using Emacs master.

I also can't seem to reproduce the "saving before the timer fires" problem either. :confused:

dgutov:
Where did the change that created a user option go?

If you like, please open a new pull request based on the current master.

PythonNut:
Dang, @dgutov is there anything that still needs doing? Sorry I was gone, it looks like you've finished it all up.

dgutov:
Yup. :)

I'm not using the new mode day-to-day (yet). So please be on a lookout for any new problems.

PythonNut:
I will. I'm using it day-to-day so hopefully things that can pop up will pop up.

PR #56, opened 2015-10-30
Define diff-hl-command-map without destroying existing bindings

tarsius:
```
Keymaps are usually defined using `defvar' which has the advantage that
it only sets a variables value if that is not defined yet.  As a result
re-evaluating (i.e. during development) a buffer that contains such a
definition does not reset the value, which would lose all user
customization.

Unfortunately `define-prefix-command' does not do that.  So use `defvar'
to define the variable and then also store the keymap in the function
cell using `fset'.
```

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -437,13 +437,14 @@ in the source file, or the last line of the hunk above it."
   (interactive)
   (diff-hl-next-hunk t))
 
-(define-prefix-command 'diff-hl-command-map)
+(defvar diff-hl-command-map
+  (let ((map (make-sparse-keymap)))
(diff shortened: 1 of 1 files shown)


dgutov:
Already merged. :)

https://github.com/dgutov/diff-hl/pull/51

tarsius:
Huh?! Seems my upstream was wrong. Sorry for da noise.

tarsius:
But I just noticed there are no tags. Could you please add one of those?

dgutov:
> Could you please add one of those?

I can. But could you clarify your use case?

Issue #57, opened 2015-11-18
Kill revert buffer on quit?


kovrik:
Currently if you call `diff-hl-revert-hunk` and answer Y or N it calls `(quit-windows-on diff-buffer)` in the end:
https://github.com/dgutov/diff-hl/blob/master/diff-hl.el#L420

This keeps `*diff-hl*` buffer open.

Consider this to be changed to `(quit-windows-on diff-buffer t)` to kill the buffer.
Alternatively, we can introduce new var `diff-hl-kill-revert-buffer-on-quit` OR new hook `diff-hl-quit-hook` (as it is done in `ediff`).

kovrik:
>you usually don't see them

If I call `diff-hl-revert-hunk` 6 times and then try to switch buffer, I see at least 6 `*diff-hl*<n>` buffers in the list.

dgutov:
Good point. So apparently, most buffer-switching commands ignore buffer ephemerity  nowadays.

I still have to ask: don't you use `vc-revert`? If it had such a custom var, `diff-hl-revert-hunk` could reuse it.

kovrik:
I can't find any var or hook in `vc-revert` that we can reuse.

So, my suggestion is to either always kill `*diff-hl*` buffer on quit or introduce new var.

dgutov:
Ah, never mind, let's just always kill it. It's not that important an issue.

kovrik:
Thanks! :)

PR #58, opened 2015-11-19
Add diff-hl-kill-revert-buffer-on-quit option

kovrik:
Introduce new variable `diff-hl-kill-revert-buffer-on-quit` (default `nil`).
If non-nil, then automatically kill `*diff-hl*` buffers on quit during `diff-hl-revert-hunk`.

This commit fixes https://github.com/dgutov/diff-hl/issues/57


dgutov:
Resolved in another way.

PR #73, opened 2016-07-18
Check that buffer is mapped to file

anmonteiro:
Certain buffers (as is the case with the Magit buffer) are not mapped to files in the file system.

Running `diff-hl-magit-post-refresh` on these buffers will cause `Wrong type argument: stringp, nil.`

It happens because I mapped `diff-hl-mode` to `prog-mode-hook`.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -538,6 +538,7 @@ in the source file, or the last line of the hunk above it."
     (dolist (buf (buffer-list))
       (when (and (buffer-local-value 'diff-hl-mode buf)
                  (not (buffer-modified-p buf))
+                 (buffer-file-name buf)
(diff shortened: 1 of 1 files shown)


dgutov:
Thanks, but have you seen the instructions?

```el
(add-hook 'prog-mode-hook 'turn-on-diff-hl-mode)
```

anmonteiro:
I didn't find that anywhere, no.
Where are those instructions?

dgutov:
The second sentence in README says:

> For the usage instructions and the list of commands, see the Commentary section inside the file.

You can also read them using `M-x describe-package`.

anmonteiro:
Gotcha, thanks

PR #79, opened 2016-10-05
Fix bug

lululau:


dgutov:
Which bug does this fix exactly?

lululau:
@dgutov `diff-hl-margin-old-highlight-function` may has a `nil` value, so if it assigns its value to `diff-hl-highlight-function` without checking if it's `nil`, it will got a `nil: void-function` error in `diff-hl-add-highlighting`.

dgutov:
> diff-hl-margin-old-highlight-function may has a nil value

When does this happen?

PR #80, opened 2016-10-23
Fix revert prompt

ryuslash:
The ‘yes-or-no-p’ procedure doesn’t pad its prompt argument with a
space so before this commit the prompt becomes:

    Revert current hunk in <file>?(yes or no)

After this commit the prompt gets fixed to:

    Revert current hunk in <file>? (yes or no)


dgutov:
Thanks!

PR #81, opened 2016-11-01
Use `user-error` when we can't proceed

Wilfred:
In these situations, it's not a bug with the code, and opening the emacs
debugger is not useful. By using `user-error`, the debugger is not
activated but we still terminate with an informative message.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -6,7 +6,7 @@
 ;; URL:      https://github.com/dgutov/diff-hl
 ;; Keywords: vc, diff
 ;; Version:  1.8.4
-;; Package-Requires: ((cl-lib "0.2"))
+;; Package-Requires: ((cl-lib "0.2") (emacs "24.3"))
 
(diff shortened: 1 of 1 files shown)


syohex:
`user-error` was introduced at Emacs 24.3. diff-hl requires Emacs 24 or higher version

dgutov:
@syohex Thanks. So this will also need `(emacs "24.3")` in the requirements section.

Should we wait until e.g. 25.2 is out?

syohex:
I don't think so. Emacs 24.1 and 24.2 are too old and there are few users. And they cannot be built with modern compiler and libraries. Emacs 25.1 is released so that it is good time to drop older Emacs support.

dgutov:
Fair enough.

Wilfred:
Okey-doke, I've updated the requirements setting.

dgutov:
Thanks!

PR #102, opened 2018-02-01
Check buffer has file name before verifying that file exists.

deadtrickster:


dgutov:
OK, I pushed a (slightly different) fix.

But if you used `turn-on-diff-hl-mode`, this problem won't have appeared at all, I think.

PR #106, opened 2018-05-09
Use an extensible generic method to determine the revision for ‘diff-…

phst:
…hl-amend-mode’


dgutov:
Umm. Do you need this to integrate a third-party backend?

dgutov:
Please let me know if you ever want to continue this conversation.

PR #109, opened 2018-07-10
Make `diff-hl-revert-hunk` work on narrowed buffers

edkolev:
This PR tweaks `diff-hl-revert-hunk` to make it work on narrowed buffers. Let me know if the PR can be improved!

Thanks for this package!

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -405,52 +405,54 @@ in the source file, or the last line of the hunk above it."
 (defun diff-hl-revert-hunk ()
   "Revert the diff hunk with changes at or above the point."
   (interactive)
-  (vc-buffer-sync)
-  (let ((diff-buffer (generate-new-buffer-name "*diff-hl*"))
-        (buffer (current-buffer))
-        (line (save-excursion
-                (unless (diff-hl-hunk-overlay-at (point))
-                  (diff-hl-previous-hunk))
-                (line-number-at-pos)))
-        (fileset (vc-deduce-fileset)))
-    (unwind-protect
-        (progn
-          (vc-diff-internal nil fileset diff-hl-reference-revision nil
-                            nil diff-buffer)
-          (vc-exec-after
-           `(let (beg-line end-line m-end)
-              (when (eobp)
-                (with-current-buffer ,buffer (diff-hl-remove-overlays))
-                (user-error "Buffer is up-to-date"))
-              (let (diff-auto-refine-mode)
-                (diff-hl-diff-skip-to ,line))
-              (save-excursion
-                (while (looking-at "[-+]") (forward-line 1))
-                (setq end-line (line-number-at-pos (point)))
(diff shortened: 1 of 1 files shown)


dgutov:
Hmm. Aren't you worried that this way the command will affect the buffer contents beyond the current narrowing?

And does the subsequent `vc-exec-after` really work fine within the narrowing?

edkolev:
> And does the subsequent vc-exec-after really work fine within the narrowing?

After some more testing - I see that it only works when reverting a hunk which adds lines. Reverting a hunk which deletes lines restores the lines in incorrect place.

Also, as you hinted, it doesn't work correctly if the hunk is partially in the narrowed region - instead of reverting the hunk, a message "Hunk hasn't been applied yet; apply it now? (y or n)" is displayed.

Do you think the following behaviour is adequate: the `diff-hl-revert-hunk`'s body could be wrapped  in a `(save-restriction (widen) ...)`. As a result, the buffer will be widened temporarily while the user is being asked "Revert current hunk in %s?".

Alternatively, I could try to achieve this behaviour: when hunk isn't 100% within the narrowed buffer, show a `(user-error "Can't revert a hunk which isn't within the narrowed buffer")`. This approach seems better from the user's stand point, but it would require a bit more changes in the code.

dgutov:
> Do you think the following behaviour is adequate

What do *you* think? You filed with issue with some particular behavior in mind, right?

> when hunk isn't 100% within the narrowed buffer, show a (user-error

That sounds okay to me, and will probably result in a shorter diff.

edkolev:
> That sounds okay to me, and will probably result in a shorter diff.

I've now implemented this behaviour, which makes the most sense to me - if the hunk is outside of the narrowed region, user-error. Otherwise, revert the hunk, preserving the narrowed region.

In case the code/behaviour can be improved - please let me know.

dgutov:
Sorry for the long wait.

After looking at the patch, it's more complicated than I would like, especially for a feature I'd like to "just work" but never test in practice.

How about we revert to the approach you mentioned?

> the diff-hl-revert-hunk's body could be wrapped in a (save-restriction (widen) ...). As a result, the buffer will be widened temporarily while the user is being asked "Revert current hunk in %s?"

I'm fine with it, and while the diff is longer, the increase in complexity is small.

@edkolev, do you have any objections against this solution yourself, as a user of narrowed buffers? I have to ask because I never use narrowing interactively.

edkolev:
> @edkolev, do you have any objections against this solution yourself, as a user of narrowed buffers? I have to ask because I never use narrowing interactively.

No, this seems fine. Reverting a hunk in a narrowed buffer isn't something that I do on a regular basis, so as long as it works correctly it's great. Having the buffer widen temporarily is perfectly fine.

I've updated the code in this PR.

dgutov:
Thanks!

PR #111, opened 2018-07-20
diff-hl-revert-hunk: show only the hunk which is about to be reverted

edkolev:


edkolev:
This PR changes the way the user is prompted whether a hunk should be revered or not.

Before this PR - the user sees all the hunk in the file:

screen shot 2018-07-20 at 19 14 54


After this PR:

screen shot 2018-07-20 at 19 15 16


My motivation for this change is that it's not clear to me which hunk is about to be reverted.

dgutov:
This needs a user option, I think. To me, it's always clear which hunk is going to be reverted, simply by the presence of the cursor.

edkolev:
I see your point. When considering terminal emacs, however, there's no cursor present, not in my setup at least (I'm not sure, maybe some state of the art linux terminal does display a cursor; I'm using iTerm and it displays no cursor):

screen shot 2018-07-23 at 10 58 23

If you feel strong about a defcustom, then please consider having the default to show only the hunk which will be reverted, without any surroundings.

Here are some options for a name:
- `diff-hl-show-full-diff-when-reverting`
- `diff-hl-narrow-diff-when-reverting`

dgutov:
You know, ideally we'd highlight the hunk-to-be-reverted somehow extra, but keep the rest visible too. Not sure which way of doing this to pick, though.

edkolev:
How about removing fontification outside of the hunk which is about to be reverted? I've pushed a commit which results in the following behavior:

screen shot 2018-08-01 at 16 25 56

I'm not sure if the code to achieve this could be considered a hack - it's basically this:
```elisp
(redisplay)
(font-lock-unfontify-buffer)
(apply #'font-lock-fontify-region hunk-bounds)
```

Note: without the call to `(redisplay)` I can't achieve the desired behaviour.

dgutov:
Hmm. That is indeed an option. Any luck with extra highlighting the current hunk, though? I was thinking more about some extra lines on both sides of it (maybe on the fringes).

dgutov:
I've tried implementing a more subtle visual cue, it's in the branch `em-reverted-hunk`.

Here's how it looks:

screenshot from 2018-11-02 01-50-38

valignatev:
Not bad at all! You keep highlighting of other changes, and the current hunk gets more visible at the same time. I like the overall idea of extra highlighting of a current hunk.

dgutov:
@valignatev Thanks!

@edkolev What do you think?

edkolev:
It's an improvement, but seems too subtle to me.

dgutov:
I should probably add a hook here, then.

edkolev:
I could implement this.

Is this what you have in mind: when reverting, a function (something like `(diff-hl-highlight-hunk beg end)`) will be called; this function will be configurable, via a radio defcustom? Should this package, diff-hl, provide a few different radio options, or just a single one?

dgutov:
@edkolev It was easier to code than to explain.

Please try out the branch I mentioned. You can now write your own highlighting function, and assign it to `diff-hl-highlight-revert-hunk-function`.

edkolev:
Great, works as expected!

For anyone interested, here are a couple of options for this new hook, I personally I'm not yet sure which I'll end up using:

```elisp
(setq diff-hl-highlight-revert-hunk-function 'evgeni-diff-hl-highlight-revert-hunk-function)

;; display only the hunk to be reverted
(defun evgeni-diff-hl-highlight-revert-hunk-function (_end)
  (diff-restrict-view))

;; or remove fontification on all hunks, except the one to be reverted
(defun evgeni-diff-hl-highlight-revert-hunk-function (end)
  (redisplay)
  (font-lock-unfontify-buffer)
  (font-lock-fontify-region (point) end))
```

dgutov:
I like the idea behind `evgeni-diff-hl-highlight-revert-hunk-function`, but because of how `diff-hl-revert-hunk` is implemented, the current hunk contains no context, and that's not ideal (to display it without any surrounding text).

dgutov:
> Great, works as expected!

Thank you for testing, `em-reverted-hunk` is now merged.

PR #112, opened 2018-07-30
Store diffs in overlays

deadtrickster:
If diffs stored in the overlay I can implement very handy tooltips like this:

diff-hl

Tooltip appears when I click on the fringe.

Code is amazingly ugly, still posting for the reference:

```lisp
(require 'pos-tip)

(global-set-key (kbd "<left-fringe> <mouse-1>")
                (lambda (ev)
                  (interactive "e")
                  (let* ((pos (posn-point (event-start ev)))
                         (overlay (car (overlays-at pos))))
                    (when overlay
                      (let* ((diff-type (overlay-get overlay 'diff-type))
                             (diff-content (overlay-get overlay 'diff-content)))
                        (when (and diff-content
                                   (or (eq diff-type 'change)
                                       (eq diff-type 'delete)))
                          (let ((lines (split-string diff-content "[\n\r]+" )))
                            (pos-tip-show-no-propertize
                             (mapconcat (lambda (line)
                                          (if (> (length line) 1)
                                              (cond
                                               ((= (aref line 0) ?+) (propertize (concat line "\n") 'face 'diff-added))
                                               ((= (aref line 0) ?-) (propertize  (concat line "\n") 'face 'diff-removed))))) (seq-drop lines 1) "")
                             'diff-added pos))))))))
```


deadtrickster:
don't worry, I'm using it for about a year, no problems so far

diff-hl.el
```diff
@@ -305,6 +306,8 @@
                     (hook '(diff-hl-overlay-modified)))
                 (overlay-put h 'diff-hl t)
                 (overlay-put h 'diff-hl-hunk t)
+		(overlay-put h 'diff-content diff)
```
dgutov: Could you avoid using tabs for indentation?

dgutov:
> don't worry, I'm using it for about a year, no problems so far

Any solid memory measurements? Like, when opening a big file.

dgutov:
Or rather, a big newly-added, or just-cleared file.

deadtrickster:
A big is how how many bytes? also how do I measure memory? RES?

deadtrickster:
btw also have button-3 bound to perform revert-hunk, very neat!

dgutov:
> also how do I measure memory? RES?

Honestly, I'd prefer to leave the method up to you. `M-x profiler-start` (in `mem` mode) could be one option.

Anyway, ideally we should extract a function one can advice here, or a couple of hooks. So whatever memory cost is there is paid only by the people who use the data.

dgutov:
> also have button-3 bound to perform revert-hunk

This is unrelated to the patch, right?

Issue #125, opened 2019-10-09
diff-hl-revert-hunk doesn't close window on success


m-cat:
First of all, thanks for this package. It's one of the most useful Emacs packages and I love `diff-hl-amend-mode`.

My issue is as the title states. Strangely, I don't have this problem unless I set `display-buffer-base-action`. Here's my full `display-buffer` config:

```elisp
      ;; Where should we open new buffers by default?
      display-buffer-base-action '(display-buffer-below-selected)
      ;; Specify custom behavior for misbehaving buffers.
      display-buffer-alist
      '(("\\*Help\\*"
         (display-buffer-reuse-window
          display-buffer-below-selected))
        ("\\*diff-hl\\*"
         (display-buffer-same-window))
        )
```

If I answer `n`, then the new window gets properly closed, but answering `y` leaves the window open (though killing the `diff-hl` buffer, so I end up with two wndows showing the same buffer):

https://www.dropbox.com/s/kqtv1zrlfzroazg/Screen%20Shot%202019-10-09%20at%2011.07.28.png?dl=0

I've been able to determine that this happens because of something in `diff-apply-hunk` -- if I comment out this function call then the new window gets properly deleted. Haven't had time to investigate this further.

Edit: forgot to mention that I haven't had this issue with any other packages or modes. I think `diff-apply-hunk` does something weird to the window configuration, preventing it from being properly cleaned up by `quit-windows-on`.

dgutov:
This will require some investigation. The `display-buffer` inner workings are pretty complex, and I'm not a pro on that subject.

If you find an easy fix, though, please let me know. You can ask at one of Emacs's mailing lists, too.

> I think diff-apply-hunk does something weird to the window configuration

The only thing I can see is that is calls `save-window-excursion`, which is fairly normal.

m-cat:
Looks like I found the cause. Commenting out this line in `diff-apply-hunk` in `vc/diff-mode.el` fixes my issue:

```elisp
      ;; Display BUF in a window
      (set-window-point (display-buffer buf) (+ (car pos) (cdr new)))
```

It seems the problem is not with `diff-hl`, but with this function in `diff-mode.el` which opens a new window containing the source buffer. (Thought this would take a while to investigate, but I found this almost immediately - the comment was a big clue).

The full fix in `diff-hl` is to add a `save-window-excursion` here:

```elisp
                (let ((diff-advance-after-apply-hunk nil))
                  (save-window-excursion
[...]

PR #127, opened 2019-10-17
Fix windows getting messed up

m-cat:
See #125

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -461,7 +461,8 @@ in the source file, or the last line of the hunk above it."
                                              ,(cl-caadr fileset)))
                   (user-error "Revert canceled"))
                 (let ((diff-advance-after-apply-hunk nil))
-                  (diff-apply-hunk t))
+                  (save-window-excursion
+                    (diff-apply-hunk t)))
                 (with-current-buffer ,buffer
                   (save-buffer))
                 (message "Hunk reverted"))))


dgutov:
What happens if the buffer visiting the file the current hunk in the diff was for is not currently visible?

I think the current code will display it in some window (seemingly a good thing), and this change will break that.

Do you have good understanding now why your display settings make windows misbehave, and how this fixes that?

m-cat:
> What happens if the buffer visiting the file the current hunk in the diff was for is not currently visible?

I would guess the file would remain not visible, but I don't know how to reproduce the scenario you mention. How is it possible to call `diff-hl-revert-hunk` on the hunk if the file is not visible?

> I think the current code will display it in some window (seemingly a good thing), and this change will break that.

If there are some important (and intended) side effects of calling `diff-apply-hunk`, they should probably be commented in `diff-hl-revert-hunk`. The only thing commented is the purpose of this command: `Revert the diff hunk with changes at or above the point.`.

If the new window is intended behavior, perhaps a configuration variable could be introduced that disables it?

I personally don't want `diff-hl-revert-hunk` making changes to my window configuration, aside from creating the temporary `*diff-hl*` buffer, whether the original buffer is visible or not (somehow).

> Do you have good understanding now why your display settings make windows misbehave, and how this fixes that?

I have no idea why this issue appears when `display-buffer-base-action` is set, and does not appear otherwise. However, it's hardly uncommon to set this variable -- I'm surprised no one's run into it before.

> how this fixes that?

`save-window-excursion` stops `diff-apply-hunk` from making any -- as far as I can tell, unintended, from the point of view of `diff-hl-revert-hunk` -- changes to the window configuration.

dgutov:
> How is it possible to call diff-hl-revert-hunk on the hunk if the file is not visible?

Fair enough!

Sorry for the late reply, merging. And thanks for the report.

PR #129, opened 2019-11-08
Fix typo

tarsius:


diff-hl.el
```diff
@@ -544,7 +544,7 @@ The value of this variable is a mode line template as in
         ;; https://github.com/magit/magit/issues/603
         (add-hook 'magit-revert-buffer-hook 'diff-hl-update nil t)
         ;; Magit versions 2.0-2.3 don't do the above and call this
-        ;; instead, but only when they dosn't call `revert-buffer':
+        ;; instead, but only when they doesn't call `revert-buffer':
```
dgutov: `don't`?

tarsius: Ah grab... :blush:

tarsius: Fixed fix.

dgutov:
Thanks!

PR #137, opened 2020-03-29
Add variable to ask for confirmation when reverting a hunk

### LINKED ISSUE #130: Remove yes-or-no-p in diff-hl-revert-hunk
It would be great if there was a way to avoid the prompt for diff-hl-revert-hunk. If I decided it was a bad idea I can always just undo it.

erickgnavar:
closes #130

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -97,6 +97,11 @@
   :group 'diff-hl
   :type 'boolean)
 
+(defcustom diff-hl-ask-before-revert-hunk t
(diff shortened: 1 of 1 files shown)


dgutov:
Thanks!

Could you amend and force-push without using tabs in indentation, though?

You can `M-x untabify`, for instance.

erickgnavar:
I haven't noticed the tabs 😅. It's fixed now

PR #140, opened 2020-05-09
Fix commands for indirect buffers.

### LINKED ISSUE #139: vc-buffer-sync: Wrong type argument: stringp, nil  in indirect buffers
Steps:
- run emacs -Q -l path/to/diff-hl.el file-under-version-control
- M-x clone-indirect-buffer
- In the new buffer, run diff-hl-diff-goto-hunk. One get the error `vc-buffer-sync: Wrong type argument: stringp, nil` with the following backtrace:

```
  file-exists-p(nil)
  vc-buffer-sync()
  diff-hl-diff-goto-hunk()
  funcall-interactively(diff-hl-diff-goto-hunk)
  call-interactively(diff-hl-diff-goto-hunk record nil)
```

- In the same buffer, vc-diff fails more gracefully with `vc-deduce-fileset: Buffer [...] is not associated with a file`

Note that the following path seems to be enough to solve the issue:
``` diff
index f504ddf..bd3caf6 100644
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -379,14 +379,15 @@ the end position as its only argument."
 (defun diff-hl-diff-goto-hunk ()
   "Run VC diff command and go to the line corresponding to the current."
   (interactive)
-  (vc-buffer-sync)
-  (let* ((line (line-number-at-pos))
-         (buffer (current-buffer)))
-    (vc-diff-internal t (vc-deduce-fileset) diff-hl-reference-revision nil t)
-    (vc-exec-after `(if (< (line-number-at-pos (point-max)) 3)
-                        (with-current-buffer ,buffer (diff-hl-remove-overlays))
-                      (diff-hl-diff-skip-to ,line)
-                      (setq vc-sentinel-movepoint (point))))))
+  (with-current-buffer (or (buffer-base-buffer) (current-buffer))
+    (vc-buffer-sync)
+    (let* ((line (line-number-at-pos))
+           (buffer (current-buffer)))
+      (vc-diff-internal t (vc-deduce-fileset) diff-hl-reference-revision nil t)
+      (vc-exec-after `(if (< (line-number-at-pos (point-max)) 3)
+                          (with-current-buffer ,buffer (diff-hl-remove-overlays))
+                        (diff-hl-diff-skip-to ,line)
+                        (setq vc-sentinel-movepoint (point)))))))

```

edit: this is GNU Emacs 27.0.50 (build 1, x86_64-w64-mingw32) of 2019-11-11

mookid:
fix #139 

and adding some tests.


Makefile
```diff
@@ -0,0 +1,24 @@
+emacs ?= emacs
```
dgutov, line 1: Could you upcase the variable names? I.e. `EMACS`, `SOURCES`, `ARTIFACTS`.

diff-hl-test.el
```diff
@@ -0,0 +1,121 @@
+;;; diff-hl.el --- Highlight uncommitted changes using VC -*- lexical-binding: t -*-
+
+;; Copyright (C) 2012-2019  Free Software Foundation, Inc.
+
+;; Author:   Dmitry Gutov <dgutov@yandex.ru>
```
dgutov, line 5: You're the author, man. :-)

diff-hl-test.el
```diff
@@ -0,0 +1,121 @@
+;;; diff-hl.el --- Highlight uncommitted changes using VC -*- lexical-binding: t -*-
+
+;; Copyright (C) 2012-2019  Free Software Foundation, Inc.
```
dgutov, line 3: I think this should be just 2020.

diff-hl-test.el
```diff
@@ -0,0 +1,121 @@
...
+
+;; Author:   Dmitry Gutov <dgutov@yandex.ru>
+;; URL:      https://github.com/dgutov/diff-hl
+;; Keywords: vc, diff
+;; Version:  1.8.7
+;; Package-Requires: ((cl-lib "0.2") (emacs "24.3"))
```
dgutov, line 9: And the rest of the headers are not needed here.

diff-hl-test.el
```diff
@@ -0,0 +1,121 @@
...
+(diff-hl-deftest diff-hl-indirect-buffer-move ()
+  (diff-hl-test-in-source
+    (narrow-to-region (point-min) (point-max))
+    (goto-char (point-max))
+    (insert "added\n")
+    (save-buffer)
+    (should-error (diff-hl-previous-hunk) :type 'user-error)
+    (should-error (diff-hl-next-hunk) :type 'user-error)))
+
+(defun diff-hl-run-tests ()
```
dgutov, line 116: I'd do this without an alias, but ok.

diff-hl-test.el
```diff
@@ -0,0 +1,121 @@
+;;; diff-hl.el --- Highlight uncommitted changes using VC -*- lexical-binding: t -*-
```
dgutov, line 1: Could you put the test file (and the test data) into a new directory? Call it `test`, I guess.

Makefile
```diff
@@ -0,0 +1,24 @@
...
+sources=
+sources+=diff-hl-amend.el
+sources+=diff-hl-dired.el
+sources+=diff-hl-flydiff.el
+sources+=diff-hl-margin.el
+
+artifacts=$(patsubst %.el, %.elc, $(sources))
+
+LOAD = -l elpa.el -l colir.el -l ivy-overlay.el -l ivy.el -l swiper.el -l counsel.el
```
dgutov, line 10: This variable seems unused.

diff-hl.el
```diff
@@ -424,56 +425,57 @@ in the source file, or the last line of the hunk above it."
 (defun diff-hl-revert-hunk ()
   "Revert the diff hunk with changes at or above the point."
   (interactive)
-  (save-restriction
```
dgutov, line 427: One way to avoid this yet-another shift of all code to the left is to create a function called `diff-hl-revert-hunk-1` and move the existing implementation into it. That might seems awkward, but the diff will be much shorter.

dgutov:
Thanks!

PR #143, opened 2020-06-18
Added option to not enable diff-hl in certain major-modes

walseb:
Currently `global-diff-hl-mode` makes `diff-hl` run in every buffer regardless of `major-mode`. This becomes a problem in `image-mode` buffers under git control in `emacs 27` built without `imagemagick` support (which is the default currently), where files ~5 mb starts taking minutes to load instead of seconds. I haven't checked to see if this also a problem when using `imagemagick` to render images, but it probably is.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -141,6 +141,28 @@ the end position as its only argument."
                  (const :tag "Highlight the first column"
                         diff-hl-revert-highlight-first-column)))
 
+(defcustom diff-hl-global-modes '(not image-mode)
+  "Modes for which `diff-hl-mode' is automagically turned on.
+This is taken directly from `font-lock-global-modes'.
+If nil, no modes have diff-hl mode automatically turned on.
+If t, all modes have diff-hl enabled.
+If a list, it should be a list of `major-mode' symbol names for which diff-hl
+mode should be automatically turned on.  The sense of the list is negated if it
+begins with `not'.  For example:
+ (not c-mode c++-mode)
+means that `diff-hl-mode' is turned on for all modes except for C and C++ mode.
+The default value disables `diff-hl' in `image-mode' buffers. Without this images
+loads slower because `diff-hl' has to look through the image data for
+possible changes. And because images are displayed on a single line in Emacs
+there is little reason to want this behavior."
+  :type '(choice (const :tag "none" nil)
+		 (const :tag "all" t)
+		 (set :menu-tag "mode specific" :tag "modes"
+		      :value (not)
+		      (const :tag "Except" not)
+		      (repeat :inline t (symbol :tag "mode"))))
+  :group 'diff-hl)
+
 (defvar diff-hl-reference-revision nil
   "Revision to diff against.  nil means the most recent one.")
 
@@ -654,8 +676,18 @@ The value of this variable is a mode line template as in
     (diff-hl-dir-mode 1))))
 
 ;;;###autoload
+(defun diff-hl--global-turn-on ()
+  "Run `turn-on-diff-hl-mode' in applicable major modes."
+  (when (cond ((eq diff-hl-global-modes t)
+	       t)
+	      ((eq (car-safe diff-hl-global-modes) 'not)
+	       (not (memq major-mode (cdr diff-hl-global-modes))))
+	      (t (memq major-mode diff-hl-global-modes)))
+    (turn-on-diff-hl-mode)))
+
+;;;###autoload
 (define-globalized-minor-mode global-diff-hl-mode diff-hl-mode
-  turn-on-diff-hl-mode :after-hook (diff-hl-global-mode-change))
+  diff-hl--global-turn-on :after-hook (diff-hl-global-mode-change))
 
 (defun diff-hl-global-mode-change ()
   (unless global-diff-hl-mode


dgutov:
Hi!

Perhaps we could model this after `font-lock-global-modes`? With the special meaning for the `not` symbol.

Another example: `company-global-modes`.

Using `'(not image-mode)` as the default sounds fine to me.

walseb:
Thanks for the suggestion! I finally got around to doing this. I just copied the mechanism directly from font-lock and it seems to work fine

walseb:
As a side note, when a image is opened it's read in fundamental-mode before switching to image-mode. This causes `diff-hl` to be run on the image data which takes a while depending on image size. This is also a problem for a lot of other minor modes and the only solution besides fixing it upstream seems to be to make them ignore fundamental-mode, which can be a problem if you edit a lot of files in fundamental-mode. So while this new option decreases the loading time of images, the global mode still slows it down a bit, at least in my testing.

diff-hl.el
```diff
@@ -141,6 +141,24 @@ the end position as its only argument."
                  (const :tag "Highlight the first column"
                         diff-hl-revert-highlight-first-column)))
 
+(defcustom diff-hl-global-ignore-modes '(not image-mode)
```
dgutov, line 144: This should be `diff-hl-global-modes`.

diff-hl.el
```diff
@@ -141,6 +141,24 @@ the end position as its only argument."
...
+  "Modes for which `diff-hl-mode' is automagically turned on.
+This is taken directly from `font-lock-global-modes'.
+If nil, means no modes have Font Lock mode automatically turned on.
+If t, all modes that support Font Lock mode have it automatically turned on.
+If a list, it should be a list of `major-mode' symbol names for which Font Lock
+mode should be automatically turned on.  The sense of the list is negated if it
```
dgutov, line 150: Not "Font Lock". :-)

diff-hl.el
```diff
@@ -141,6 +141,24 @@ the end position as its only argument."
...
+begins with `not'.  For example:
+ (c-mode c++-mode)
```
dgutov, line 152: You can probably use a list `(prog-mode)` as a better example. Nobody's going to enable this is C and C++ only.

diff-hl.el
```diff
@@ -653,9 +671,19 @@ The value of this variable is a mode line template as in
    ((eq major-mode 'vc-dir-mode)
     (diff-hl-dir-mode 1))))
 
+;;;###autoload
+(defun diff-hl--global-turn-on ()
+  "Run `turn-on-diff-hl-mode' if current major mode isn't ignored."
```
dgutov, line 676: ...in applicable major modes.

dgutov:
> Nevermind, I was going to report this upstream but now I can't replicate it. I must have done something wrong in the earlier tests

I'm guessing this might have been a difference between just visiting a file for the first time and `M-x image-mode` in an existing buffer. But I'm not sure.

diff-hl.el
walseb: So the font-lock implementation only looks at concrete modes like C++ and C mode because it just uses `(memq major-mode diff-hl-global-ignore-modes)`. 

I have implemented the described behavior here: `(seq-find (lambda (a) (derived-mode-p a major-mode)) diff-hl-global-modes)`, but the problem is that using `seq-find` requires emacs 25.1 while this package currently support 24.4

Do you think it's ok to bump the requirements up to emacs 25.1? Is anyone still using 24.4? Or should I work around this by defining a custom `seq-find` implementation?

walseb:
Do you think it would also be ok to add `minibuffer-inactive-mode` to the default ignore list? Is there any use case where you would want `diff-hl` in the minibuffer? I haven't observed any downsides to running it in the minibuffer however, so maybe this isn't worth doing

diff-hl.el
dgutov: > So the font-lock implementation only looks at concrete modes like C++ and C mode because it just uses (memq major-mode diff-hl-global-ignore-modes).

I'm talking about the example in the documentation. Not the implementation.

> I have implemented the described behavior here: (seq-find (lambda (a) (derived-mode-p a major-mode))

This is not actually how `derived-mode-p` is used. You can try this: `(apply #'derived-mode-p diff-hl-global-modes)`.

OTOH, if anyone wanted to disable `diff-hl` in `fundamental-mode` only, perhaps this is not a good idea. Because every major mode is derived from fundamental. Then what `turn-on-font-lock-if-desired` is doing is better. IDK which one to pick, honestly, but in `company-mode-on` I just followed its example.

dgutov:
> Do you think it would also be ok to add minibuffer-inactive-mode to the default ignore list?

No need. `buffer-file-name` is nil in minibuffer.

diff-hl.el
walseb: Yeah I agree I think it's probably better to keep it like it is in `company-mode` and `font-lock`. Maybe to make the example more realistic it should be changed to `(not c-mode c++-mode)` for example

dgutov: Perhaps just explain the default value in the documentation (that there's little point is using this mode for images, and it was too slow as well, and the default value avoids it).

walseb:
I have now merged all the commits and fixed all the problems I think

dgutov:
Looks good. I'll edit it a bit later to make it shorter.

Thanks!

Issue #134, opened 2020-02-02
Generalize diff-hl-amend-mode?


AmaiKinono:
Having this mode is great, but sometimes we want to show changes against some early version. Like when working on a PR, there are a bunch of commits and we want to see what the whole PR changes. Could we ask the user to specify the reference commit (by input a number or interactively choose from say last 15 commits)?

dgutov:
We have a variable you can set (the same one that `diff-hl-amend-mode` uses). You could just `setq` it.

As far as adding a UI to change that var interactively... well, it should ideally be backend-agnostic. If we just ask a Git user to choose between 15 commit hashes, that wouldn't be very user-friendly.

AmaiKinono:
Thanks. I just saw `diff-hl-reference-revision`, and that does the work.

I tried to experiment with the UI. There's a `vc-read-revision` offerd by `vc.el`, and seems we can use that. But then I found for git, it only shows tagged revisions and branches. And [here](https://www.gnu.org/software/emacs/manual/html_node/emacs/VCS-Concepts.html#VCS-Concepts) it says:

> Each commit creates a new revision in the repository.

I don't know `vc.el` well. Do you have any idea why this happens or does it work with other VCS backends?

dgutov:
> But then I found for git, it only shows tagged revisions and branches.

Because there is little point in completion for SHA hashes. You can paste one if you like, though. But just as well you can `(setq diff-hl-reference-revision "rev-hash")`.

> Do you have any idea why this happens

What happens? The manual just described how all version control systems work.

AmaiKinono:
I think the manual says a "revision" means a commit in git, so I should see all commits when calling `diff-hl-reference-revision`.

> Because there is little point in completion for SHA hashes.

This makes sense. I will see if I can make use of `vc-print-log` instead.

dgutov:
Yeah, revision and commit are the same in the context of Git.

dgutov:
BTW, when working on a PR, you might be working against some other branch, such as `master` or `develop`. Symbolic references like that are also allowed where a revision is expected.

dgutov:
And merged. Thanks!

PR #146, opened 2020-10-29
Set reference rev in change log buffer

AmaiKinono:
Solves #134.

Usage: 

- `M-x vc-print-log`, then run `diff-hl-log-view-set-reference-rev-for-fileset` on a rev to set it as the reference for the fie. Run `diff-hl-reset-reference-rev-for-fileset` to reset.
- `M-x vc-print-root-log`, then run `diff-hl-log-view-set-reference-rev-for-repo` on a rev to set it as the reference for all files in the repo. Run `diff-hl-reset-reference-rev-for-repo` to reset.

I'll write the usage in README after we are done with the code.

Problems:

- I am not familiar with the `fileset` concept of `vc`. To me it's like always being the current file. If this is right, I'll change the command name and docstring to make more sense for the users.
- For some reason, after setting the reference rev, I have to disable and enable `diff-hl-mode` in the file buffers for it to take effect.
- Set reference rev for repo doesn't affect newly opened files in that repo. Maybe we could set up some hook to do this.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -55,6 +55,8 @@
 (require 'diff-mode)
 (require 'vc)
 (require 'vc-dir)
+(require 'log-view)
+
 (eval-when-compile
   (require 'cl-lib)
   (require 'vc-git)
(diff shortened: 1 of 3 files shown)


dgutov:
A fileset can be multiple files if you are in a VC-Dir or Dired buffer with several marked entries. It can also be a directory if the current buffer doesn't correspond to a single file but a whole repository (e.g. the result of `M-x vc-print-root-log`).

This seems a bit too complex to me. Surely you could do with code that just changes the `diff-hl-reference-revision` variable globally. Did that seem insufficient, and you wanted the change to only affect a single repository?

Per-file changes like that can be sneaky. I think it should be easy to forget that a given buffer is using a different value of this variable. When the value is global, it's much less likely to be forgotten.

dgutov:
> For some reason, after setting the reference rev, I have to disable and enable diff-hl-mode in the file buffers for it to take effect

Does making a change and then saving a buffer suffice as well?

Or `diff-hl-log-view-set-reference-rev-for-fileset` could call `diff-hl-update` at the end.

AmaiKinono:
> Per-file changes like that can be sneaky. I think it should be easy to forget that a given buffer is using a different value of this variable. When the value is global, it's much less likely to be forgotten.

Reasonable. Please see if you are happy with the new version.

dgutov:
Looking good.

How do you feel about copyright assignment to FSF? `diff-hl` in distributed via GNU ELPA, and they/we have that requirement.

The patch, even as reduced as the new version is, is still above the allowed threshold for non-assigned contributions.

AmaiKinono:
Oh, I didn't notice that `diff-hl` is in GNU elpa. I am ok with the copyright assignment. I'll start it soon.

dgutov:
Excellent.

In case you were wondering where to start, the instructions are here: https://code.orgmode.org/bzg/org-mode/raw/master/request-assign-future.txt

AmaiKinono:
I've got my assignment done. What are the next steps?

dgutov:
A NEWS entry would be nice.

Also, here's an idea: `diff-hl-log-view-set-reference-rev` doesn't really *have* to always be about log-view buffers. It could work in different kinds of buffers, and suggest to users revisions based on each contexts. When called in a log-view, it could suggest the value returned by `(log-view-current-tag)`. When called in a `vc-annotate` buffer (if you like), it could pick up the revision on the current line. And otherwise, just look around point and see whether there's something resembling a revision there.

I think it should be called `diff-hl-set-reference-rev`, take a revision as an argument, and its `interactive` form can do auto-detection described above.

AmaiKinono:
Great idea. I can imagine people use this in a terminal/shell buffer.

I've changed the command to what you described. One problem is I don't know how to decide if a revision string is valid or not, so if people call it by accident on a random symbol, they may end up having a invalid reference revision.

> A NEWS entry would be nice.

I haven't found a NEWS file or section in the README.md. Do we want to add one now?

dgutov:
> I haven't found a NEWS file or section in the README.md. Do we want to add one now?

Oh, sorry, I was thinking of a different project. It's fine now.

> One problem is I don't know how to decide if a revision string is valid or not, so if people call it by accident on a random symbol, they may end up having a invalid reference revision.

That's possible, I guess. But they will have to confirm it with RET first. Unfortunately, there is no backend action to verify a revision. Anyway, that's a future problem.

diff-hl.el
```diff
@@ -683,6 +687,51 @@ The value of this variable is a mode line template as in
...
+  "Set the reference revision globally to REV.
+When called interactively, REV is get from contexts:
+
+- In a log view buffer, it uses the revision of current entry.
+Call `vc-print-log' or `vc-print-root-log' first to open a log
+view buffer.
+- In a VC annotate buffer, it uses the revision of current line.
+- In other situations, get the revision name at point.
+
+Notice that this sets the reference revision globally, so in
```
dgutov, line 701: Maybe if this turns out to be a real problem, the "reference revision" structure could evolve into an alist, mapping repository roots to reference revisions.

PR #147, opened 2020-11-08
Use posframe or popup to show the current hunk

alvarogonzalezsotillo:
This PR is a follow up of https://github.com/dgutov/diff-hl/pull/145

I refactored the code to allow several backends to to show the differences of the current hunk. Currently, `posframe` and `popup` are implemented. Also, there is a mode `diff-hl-show-hunk-mode` with new key bindings (`C-v x *`, `C-v x {` and `C-v x }`) and another mode `diff-hl-show-hunk-mouse-mode` to enable the mouse in the fringe and the margin.

`popup` seems to work fine from the begining. `posframe` has been harder, but the current version works with Ubuntu (KDE, Gnome 3) and Windows.

I would like to include `diff-hl-show-hunk` in the `diff-hl-command-map` as `C-x v *`, but It would require a dependency from `diff-hl` to `diff-hl-show-hunk`, instead of the other way arround (not very clean). Do you think it will be possible in any other way?

I made minor changes in `README.md`, but I can add more instructions (see https://github.com/alvarogonzalezsotillo/diff-hl-show-hunk#customization)

The images added to the readme file have not the same style as the previous ones. Maybe you can get a static image with your own settings and change them. If not, I can change my theme in emacs and take again all the images.

> Finally, this package is part of GNU ELPA, so to accept a contribution like this I'll have to ask you to complete copyright assignment to FSF. Would you be willing to do that?

If you like this PR, I will need some directions about how to proceed with the FSF.


diff-hl-show-hunk-posframe.el
```diff
@@ -159,6 +159,9 @@ to scroll in the posframe")
 
   (set-frame-parameter diff-hl-show-hunk--frame 'drag-internal-border t)
   (set-frame-parameter diff-hl-show-hunk--frame 'drag-with-header-line t)
+
+  ;; Dedicate window
+  (set-window-dedicated-p (window-main-window diff-hl-show-hunk--frame) t)
```
dgutov, line 164: Seems like posframe already does this in `posframe--create-posframe`.

dgutov:
Hi! Sorry for the late reply.

I like this in principle, but we'll need to discuss some details. And also it would be nice to have a review from the target audience (perhaps @deadtrickster would like to take a look?).

> I would like to include diff-hl-show-hunk in the diff-hl-command-map as C-x v *, but It would require a dependency from diff-hl to diff-hl-show-hunk, instead of the other way arround (not very clean).

Do you want them there always or only when one of the new minor modes is enabled?

> The images added to the readme file have not the same style as the previous ones. Maybe you can get a static image with your own settings and change them. If not, I can change my theme in emacs and take again all the images.

Yeah, the current screenshots here don't reflect the theme I'm using now either. I can redo them. Or we could standardize on the default Emacs theme. No dark themes, please. Though we could add some extra (dark) screenshots in a directory inside, and link to them.

> If you like this PR, I will need some directions about how to proceed with the FSF.

Please see these instructions: https://code.orgmode.org/bzg/org-mode/raw/master/request-assign-future.txt

(Keep the answer to the first question that's written there, "Emacs"). That will let you contribute to Emacs itself later, as well as other packages in GNU ELPA.

### Other notes

Two new modes is a good, conservative choice, considering some other feature might want to use the fringe/margin as well, and there is no existing way for such features to cooperate.

Since `popup` and `posframe` are not `require`-d by any of the features here (except at runtime), we probably need a bunch of `declare-function` declarations to clean up the byte compilation warnings.

Whenever you have several empty lines in a row (in the code), you can replace them with just one.

The new files need the appropriate license and copyright blurbs, as well as the author header (if it stumps you, I can add those later).

What about the default width? Will it often be shorter than the diff inside? If we're emulating a feature from some other editor, could you link to some screenshot, so we can compare the behavior?

On the screenshots, I see the popups showing more than just one hunk. Is that intentional? I wonder what the users prefer in general.

diff-hl-show-hunk-posframe.el
alvarogonzalezsotillo: You are right, I will undo this change

deadtrickster:
Awesome, I will try it soon!

alvarogonzalezsotillo:
> > I would like to include diff-hl-show-hunk in the diff-hl-command-map as C-x v *, but It would require a dependency from diff-hl to diff-hl-show-hunk, instead of the other way arround (not very clean).
> 
> Do you want them there always or only when one of the new minor modes is enabled?
>
I would like to be always available. Anyway, the user can add its own keybindings, so it is not very important.

 
> Yeah, the current screenshots here don't reflect the theme I'm using now either. I can redo them. Or we could standardize on the default Emacs theme. No dark themes, please. Though we could add some extra (dark) screenshots in a directory inside, and link to them.
I agree dark themes are _opinionable_ :) 
Even if I use the default theme, I have not the same default font, dpi  and window decorations. Can you take the new screenshots? I think is the best solution, and I can repeat the animated gifs and put them in a "screencast" directory.



> Two new modes is a good, conservative choice, considering some other feature might want to use the fringe/margin as well, and there is no existing way for such features to cooperate.
> 
> Since `popup` and `posframe` are not `require`-d by any of the features here (except at runtime), we probably need a bunch of `declare-function` declarations to clean up the byte compilation warnings.
I will investigate this `declare-funciton` thing (I am still pretty new in emacs-lisp). I supose there will be something link `declare-variable` also.

 
> Whenever you have several empty lines in a row (in the code), you can replace them with just one.
> 
Ok

> The new files need the appropriate license and copyright blurbs, as well as the author header (if it stumps you, I can add those later).
Ok
 
> What about the default width? Will it often be shorter than the diff inside? 
I don't change the width of the posframe, it seems to adjust automatically to the contents, and sometimes the width is smaller that the contents.
There is a custom variable `diff-hl-show-hunk-posframe-poshandler` that is a function that changes the posframe position and size. I can try to use some function like `string-width` to calculate the max line width, and change the width of the posframe.

> If we're emulating a feature from some other editor, could you link to some screenshot, so we can compare the behavior?
I think I saw something like this in eclipse, called quick diff
image


> On the screenshots, I see the popups showing more than just one hunk. Is that intentional? I wonder what the users prefer in general.
The content of the popup is determined by `diff-hl-show-hunk-narrow`. It can be:
- `nil`: The complete diff 
- `t`: The _hunk_ that is contained between two `@@......@@` lines. This depends on the diff algorithm used by the underlying VCS (I suppose, only tested in git), and I think is the part of the file that will be reverted if `diff-hl-revert-hunk`

dgutov:
> I would like to be always available. Anyway, the user can add its own keybindings, so it is not very important.

Perhaps it should be added (and removed on disable) by `diff-hl-show-hunk-mode`?

> Even if I use the default theme, I have not the same default font, dpi and window decorations. Can you take the new screenshots? I think is the best solution, and I can repeat the animated gifs and put them in a "screencast" directory.

Yes, sure. We'll do that at the end, after everything else is taken care of.

> I supose there will be something link declare-variable also.

The counterpart for variables is just `defvar`, but without INITVALUE. It can be at the top level, or only inside the function that needs it (usually at its beginning).

> I don't change the width of the posframe, it seems to adjust automatically to the contents, and sometimes the width is smaller that the contents.

When the diff is narrower than the current window, I suppose it's okay. But if it's wider, it would be better to use the width of the current window, I think.

Either way, perhaps we can do without `diff-hl-show-hunk-popup-default-width`? Or do you consider this customization valuable enough?

> The content of the popup is determined by diff-hl-show-hunk-narrow.

Ok.

> and I think is the part of the file that will be reverted if diff-hl-revert-hunk

Not exactly. Or, at least, while it is functionally the same, `diff-hl-revert-hunk-1` makes an effort to separate the actual change from the context (by splitting it twice, usually). I wonder whether the users might prefer this, considering otherwise the context should be apparent in the parent buffer. Or not.

> I think I saw something like this in eclipse, called quick diff

So Eclipse was your inspiration?

diff-hl-show-hunk-popup.el
```diff
@@ -1,10 +1,50 @@
 ;;; diff-hl-show-hunk-popup.el --- popup backend for diff-hl-show-hunk -*- lexical-binding: t -*-
 
+;; Copyright (C) 2020  Free Software Foundation, Inc.
+
+;; Author:   Dmitry Gutov <dgutov@yandex.ru>
+;;           Álvaro González <alvarogonzalezsotillo@gmail.com>
```
dgutov, line 6: I think it would be more correct to put just you as the author of this and other files here.

diff-hl-show-hunk-popup.el
```diff
@@ -1,10 +1,50 @@
...
+;; URL:      https://github.com/dgutov/diff-hl
```
dgutov, line 7: The URL header is unnecessary here, it's only read from the main .el file by the package.el infrastructure.

diff-hl-show-hunk-popup.el
```diff
@@ -1,10 +1,50 @@
...
+(declare-function popup-scroll-up "popup")
+(declare-function popup-height "popup")
+(declare-function popup-scroll-down "popup")
+(declare-function popup-hide "popup")
+(declare-function popup-hide "popup")
+(declare-function popup-create "popup")
+(declare-function popup-set-list "popup")
+(declare-function popup-select "popup")
+(declare-function popup-draw "popup")
+(eval-when-compile
```
dgutov, line 40: `eval-when-compile` is unnecessary here, IIRC.

RosanneZe:
I've been trying out the feature and I like it. One point of feedback I would have for the popup version (I still need to install posframe and try it out), is that it would be good if the popup could always popup at the start of the line, so the aligment of the diff in the popup is the same of that of the file.

deadtrickster:
I tried it out today, couldn't make posframe integration work (likely I messed up with my cache because I (mis)use gccemacs). One thing I spotted immediately - when I click on fringe it asks to save file. In my version I specifically avoided saving, I think it is not convenient and error prone. How to turn this off?

alvarogonzalezsotillo:
> I tried it out today, couldn't make posframe integration work (likely I messed up with my cache because I (mis)use gccemacs). > One thing I spotted immediately - when I click on fringe it asks to save file. In my version I specifically avoided saving, I think it is not convenient and error prone. How to turn this off?

I use `diff-hl-diff-goto-hunk` to find the differences, that calls `vc-buffer-sync`, that saves the buffer. I don't know if there is other way to get the differences.

If I get this PR merged, my plan is to store the diffs in the overlays (https://github.com/dgutov/diff-hl/pull/112) as an option, with a customize flag like `diff-hl-store-diffs-in-overlays`, and use `diff-hl-diff-goto-hunk` or the overlays based on that customized variable.

alvarogonzalezsotillo:
>  is that it would be good if the popup could always popup at the start of the line, so the aligment of the diff in the popup is the same of that of the file.

Now I use `point-at-bol` to show the popup

> Either way, perhaps we can do without diff-hl-show-hunk-popup-default-width? Or do you consider this customization valuable enough?

I used a customized variable because I didn't know what value to use. But now, it seems clear that the correct width for the popup is the full width, so i removed `diff-hl-show-hunk-popup-default-width`.

Still working in the width of the posframe.

> So Eclipse was your inspiration?

I used Eclipse long time ago, and I liked quick-diff, and #112 was the final inspiration.
But the _motivation_ is my son, that uses VSCode, and I need to show him that emacs and I are still relevant :)

> On the screenshots, I see the popups showing more than just one hunk. Is that intentional? I wonder what the users prefer in general.

I think the solution could be the *contex* of the diff algorithm. Two hunks can be overlaped in the output of `vc-diff-internal` if the context arround the changes overlap.

If I customize `vc-git-diff-switches`  to `--unified=0` the hunks remain isolated.

For SVN, maybe setting `vc-svn-diff-switches` to `--diff-cmd=diff -x -U0` could work.

RosanneZe:
> Now I use point-at-bol to show the popup
I'm still using your repo version, so I don't know if you updated this, but for me, if I mouse click on the fringe to get the popup (configured at the right fringe), it pops up at the end of line. 
If I do M-x diff-hl-show-hunk the popup starts at the point.
But if you're going to make the popup full width, this might not be an issue anymore.

deadtrickster:
> If I get this PR merged, my plan is to store the diffs in the overlays (#112) as an option, with a customize flag like `diff-hl-store-diffs-in-overlays`, and use `diff-hl-diff-goto-hunk` or the overlays based on that customized variable.

why not to make overlays default?

dgutov:
> why not to make overlays default?

Perhaps that would not be optimal for both possible values of `diff-hl-show-hunk-narrow`? When it's t, saving a hunk to an overlay should work, but not in the other case.

dgutov:
And either way, the code could reuse the method employed by `diff-hl-flydiff`, though it might require a bit more effort.

dgutov:
@alvarogonzalezsotillo

> Now I use point-at-bol to show the popup

Very good.

> I used a customized variable because I didn't know what value to use. But now, it seems clear that the correct width for the popup is the full width, so i removed diff-hl-show-hunk-popup-default-width.

Yeah, looks like full width is the best. VS Code uses the full width anyway: https://stackoverflow.com/a/47296478/615245

> But the motivation is my son, that uses VSCode, and I need to show him that emacs and I are still relevant :)

One of the best motivations! Props to you.

> I think the solution could be the contex of the diff algorithm.

Right. Check out the macro `diff-hl-with-diff-switches`, it already applies said switches in a way that works for several VCSes. That's how `diff-hl-changes-buffer` gets the isolated hunks.

alvarogonzalezsotillo:
> > why not to make overlays default?
> 
> Perhaps that would not be optimal for both possible values of `diff-hl-show-hunk-narrow`? When it's t, saving a hunk to an overlay should work, but not in the other case.

I am afraid `diff-hl-show-hunk-narrow` is here for historical reasons: my initial code didn't extract the hunk, because I didn't know how to do it. At this moment, I think there is no reason to maintain this option, and narrow-to-hunk is always the good choice.

If there are no objections, I will remove it.

deadtrickster:
Fully understanding no one owes nothing to nobody, I would still like to point out that forcing save might be a show stopper for others as well. At least in my workflow I reach out to diffs when I think I'm screwing up, other users could have those live-reload-on-save a-la node.js systems and forcing users to save might be undesirable.

dgutov:
@alvarogonzalezsotillo 

> If there are no objections, I will remove it.

No objections. But wouldn't that also obviate the need for buttons in the posframe popup? Or do they still play a purpose, when the current hunk is long?

@deadtrickster

The `diff-hl-flydiff` method I mentioned shouldn't require the file to be saved. See `diff-hl-flydiff-buffer-with-head`.

Putting hunks inside the overlays is an option, but I didn't like it because they still would be there even if this mode is unused. But it's definitely an option which we can go back to if the other approaches don't work to our satisfaction.

alvarogonzalezsotillo:
> Fully understanding no one owes nothing to nobody, I would still like to point out that forcing save might be a show stopper for others as well. At least in my workflow I reach out to diffs when I think I'm screwing up, other users could have those live-reload-on-save a-la node.js systems and forcing users to save might be undesirable.

I agree, but in this first "version" I'm focused on make it work and get integrated in diff-hl. 

The next step could be to implement a version of `diff-hl-diff-goto-hunk` using `diff-hl-flydiff-buffer-with-head` to get the diffs instead of `vc-diff-internal`, maybe with an advice when `diff-hl-flydiff-mode` is enabled.

alvarogonzalezsotillo:
> But wouldn't that also obviate the need for buttons in the posframe popup? Or do they still play a purpose, when the current hunk is long?

The buttons can be used to navigate to other hunks, and contribute to the "discoverability" of some commands. They can be disabled with `diff-hl-show-hunk-posframe-show-header-line`.

dgutov:
> The buttons can be used to navigate to other hunks, and contribute to the "discoverability" of some commands

Makes sense. The buttons are pretty cool anyway.

> They can be disabled with diff-hl-show-hunk-posframe-show-header-line.

I was also thinking maybe it some later version they could migrate to the side (either left or right), and thus take less screen space, and giving even less reasons to disable them.

And if/when someone implements the second-best feature request (showing the indicators against the staging area), some button could toggle the hunks base between the working revision and the staging area. If the difference in "current hunks" doesn't torpedo that idea, though.

alvarogonzalezsotillo:
> Fully understanding no one owes nothing to nobody, I would still like to point out that forcing save might be a show stopper for others as well. At least in my workflow I reach out to diffs when I think I'm screwing up, other users could have those live-reload-on-save a-la node.js systems and forcing users to save might be undesirable.

See https://github.com/dgutov/diff-hl/pull/151 . This would allow the posframe/popup without saving the file.

dgutov:
@alvarogonzalezsotillo Sorry if I was unclear. This approach should work fine whether `diff-hl-flydiff-mode` is enabled or not (right?), so instead of creating a new advice, I think it should be easy enough to extract a helper function from that package into the `diff-hl` package proper and use it from both `diff-hl-flydiff-mode` and the new feature.

As the first approximation, you can even autoload `diff-hl-flydiff-buffer-with-head` and use it directly (I can take a look later at whether further reorganization is really necessary).

Also, instead of calling `diff-hl-diff-goto-hunk` itself, you can create a specialized version of that function just for use in `diff-hl-show-hunk` that would call `diff-hl-flydiff-buffer-with-head` instead of `vc-diff-internal`. It wouldn't need to call `diff-hl-remove-overlays`, and it would use a direct control flow without `vc-exec-after` because `diff-hl-flydiff-buffer-with-head` uses `'noasync`.

Please let me know if you have any questions about this proposal.

dgutov:
@alvarogonzalezsotillo That's exactly right. Sorry for the late reply.

alvarogonzalezsotillo:
Please check the commit https://github.com/dgutov/diff-hl/pull/147/commits/f97b216a47c312aa81db082525acb8b8c53ea0ae . I think it contains the changes you requested.

There could be a potential problem with emacs versions <= 25, since there is a check in `diff-hl-flydiff.el` that i dont fully understand

      ;; Polyfill concrete revisions for vc-git-working-revision in Emacs 24.4, 24.5
      (when (version<= emacs-version "25.0")
         (with-eval-after-load 'vc-git
            (advice-add 'vc-git-working-revision :override
                #'diff-hl-flydiff/vc-git-working-revision)
            (advice-add 'vc-git-mode-line-string :override
                #'diff-hl-flydiff/vc-git-mode-line-string)))
 
These two advices (and the corresponding functions) should be moved to `diff-hl.el`?

alvarogonzalezsotillo:
Please forget f97b216 .

When I restarted emacs it doesn't work. It seems i evaled some code that worked, and commited another one.

I'm working on int.

dgutov:
> There could be a potential problem with emacs versions <= 25, since there is a check in diff-hl-flydiff.el that i dont fully understand

Yeah, I'll probably just bump the minimum version number when we merge this.

dgutov:
So don't worry about it.

alvarogonzalezsotillo:
I think the  commit 70debb0 solved the problem in f97b216.

diff-hl.el
```diff
@@ -664,6 +664,80 @@ The value of this variable is a mode line template as in
...
+                 file
+                 (or diff-hl-reference-revision
+                     (diff-hl-working-revision file)))))
+      ;; FIXME: When against staging, do it differently!
+      (diff-no-select rev (current-buffer) "-U 0 --strip-trailing-cr" 'noasync
+                      (get-buffer-create dest-buffer))
+      (with-current-buffer dest-buffer
+        (let ((inhibit-read-only t))
+          ;; Function `diff-sentinel' adds a final line, so remove it
+          (delete-matching-lines "^Diff finished.*")))
```
dgutov, line 738: Hey, it's looking pretty good.

Could you explain this bit, though? If we're just showing the current hunk, the "final line" should not matter, should it?

diff-hl-show-hunk.el
```diff
@@ -0,0 +1,326 @@
...
+
+;; `diff-hl-show-hunk' shows a posframe/popup with the modified hunk at point.
+;; `diff-hl-show-hunk-function' contains the backend used to show the hunk.  Its
+;; default value is `diff-hl-show-hunk-function-default', which tries to use
+;; `diff-hl-show-hunk-posframe' (GUI), and then `diff-hl-show-hunk-popup'.
+;; Other backends (for example pos-tip or phantom overlays) could be
+;; implemented.
+;;
+;; It uses the same diff algorithm as `diff-hl-flydiff-mode', so it is
+;; highly recomended to enable `diff-hl-flydiff-mode' before use this package.
```
dgutov, line 32: Also about this paragraph.

I'd say it's phrased too strongly. I mean, `diff-hl-flydiff-mode` is good and all, but IIUC there won't be any serious cases when you pop up the current hunk and see something unexpected (e.g. a hunk that is different). Will there be?

Because if the original logic in `diff-hl` the hunks that are still highlighted on the fringe (or margin) must all be up-to-date.

diff-hl-show-hunk.el
```diff
@@ -0,0 +1,326 @@
...
+;;
+;;    ```
+;;      (global-diff-hl-show-hunk-mode)
+;;    ```
+;;
+;; Two hunks can overlap due to the contex of the diff algorithm.  To
+;; avoid this overlap, set 0 contex lines in `vc-diff'.  The following
+;; backends are automatically configured:
+;;
+;; - GIT: `vc-git-diff-switches' uses '--unified=0'.
```
dgutov, line 47: You document these switches here, but they are actually reused from the existing code in `diff-hl`, right?

diff-hl.el
alvarogonzalezsotillo: The `diff-sentinel` function is called internally from `diff-no-select`. At the end of the function, a line with `Diff finished (status) (date)` is added to the output of the `diff` command. This line appeared in the posframe/popup of the last change in the buffer, but not in the other popups. 

I removed it for maintain consistency.

diff-hl-show-hunk.el
alvarogonzalezsotillo: You are right, this is an outdated comment of a previous commit (before implementing the `diff-hl-diff-buffer-with-head` thing).
I am removing it.

diff-hl-show-hunk.el
alvarogonzalezsotillo: I was worried about inconsistencies between the diffs showed in the margin and the diffs in the popups.
Maybe It is phrased too strongly (sorry, english is not my best language). I am removing this comment about `diff-hl-flydiff-mode`.

alvarogonzalezsotillo:
I think I have addressed all the problems in aea3301. BTW, my assignment/disclaimer process with the FSF is complete.

I have also developed a third backend to show the diffs: `inline-popup.el`. I think `diff-hl-show-hunk-inline-popup` should be the "new" default backend:

- It has no external dependencies
- It works well in GUI and terminal
- It feels faster that posframe, at least in my pc
- It is prettier (in my opinion) than popup

inlup-600

There is a problem, anyhow:
- It has a problem that I cant handle: when using `text-scale-adjust`, I am not able to draw the horizontal lines with the visible width of the window.

dgutov:
This looks very nice. Perhaps we'll even replace the `popup` backend with it, if it's always better?

> when using text-scale-adjust, I am not able to draw the horizontal lines with the visible width of the window

There is a handy package for changing page breaks into horizontal lines, and here is the list of hooks it uses:

https://github.com/purcell/page-break-lines/blob/34cd91d841e5334c6966e25fbfd971a2dfbcee69/page-break-lines.el#L111-L115

alvarogonzalezsotillo:
> There is a handy package for changing page breaks into horizontal lines

I already have inspected it searching for inspiration.

It uses the  face `page-break-lines` for the line separator, and that face doesn't change size with `text-scale-adjust` (I don't know why). It is self evident if you change `page-break-lines-char` with another char (`o`, for example).

> Perhaps we'll even replace the popup backend with it, if it's always better?

Thats also my opinion. I can make an independent package when this PR get merged, just in cases someone find it useful.

dgutov:
> It uses the face page-break-lines for the line separator, and that face doesn't change size with text-scale-adjust (I don't know why). It is self evident if you change page-break-lines-char with another char (o, for example).

Do you mean the vertical size? If so, that seems fine to me. After all, its only purpose is to draw a thin line, not a rectangle.

alvarogonzalezsotillo:
This video shows that the width and height of the break-line characters doesn't change. I suppose this can be done only changing the glyph of the display table, and it is not possible with overlays.

pagebreak

Although this is a problem, I think it is not very important, and the package can be released as is.
Another (but ugglier solution) is to align the text to the left instead.

alvarogonzalezsotillo:
I found the way to get the correct character size (https://emacs.stackexchange.com/questions/5495/how-can-i-determine-the-width-of-characters-on-the-screen). Now, the width of the inline popup fits the width of the window.

I have made `diff-hl-show-hunk-inline-popup` the default function. Posframe and popup are still available, but not loaded by default

dgutov:
Thanks, I'll give it a try this weekend. After cutting a release with the current code, so we'll have some time to experiment before the next one.

BTW, to satisfy some formalities, I think the MELPA folks frown on abbreviations like `inlup`, so writing the package name in full inside the function names would be best (if `exec-path-from-shell` does that, we can do it too ;-)).

In addition to that, I'll have to ask whether it's okay to keep a differently named package in the same repo, or if we should rename it to start with `diff-hl-`. We could publish it separately, though. If you think it will be useful independently.

alvarogonzalezsotillo:
> I think the MELPA folks frown on abbreviations like inlup

I developed the `region-occurrences-highlighter`package, and I didn't like the long name, so I tried a shorter name, but I agree `inlup`it is not very descriptive. I have changed `inlup*` for `inline-popup`.

> We could publish it separately, though. If you think it will be useful independently.

Yes, I think so. Maybe it can be integrated in `diff-hl` in a first version, and then separate it, since I'm not very confident about the usefullness/stability of my own inline popup. For example, I don't know if setting the parent keymap of a minor-mode in each invocation of `inline-popup-show` can be considered hacky and fragile, or not.

dgutov:
Hi!

In lieu of a review, I've now pushed your branch with a number of changes and comments to this repository, branch with the same name: https://github.com/dgutov/diff-hl/tree/diff-hl-show-hunk

Check it out. Also, if you think we still need some of the things I've removed, please let me know.

I haven't looked at `inline-popup`'s code yet, that's still coming.

alvarogonzalezsotillo:
I have added some comments on your commits, but I agree with the changes in https://github.com/dgutov/diff-hl/tree/diff-hl-show-hunk.

dgutov:
Thanks. Overall, I like how it works.

Having tried it out for a while, I have just one concern: popup positioning. Don't you think that both the posframe popup and the inline one should be aligned vertically with the indicator on the fringe? In practical terms, it would mean always showing the popup in the column zero, and on the line where the hunk overlay begins (covering it).

Right now, the posframe popup especially looks kinda random, showing up wherever the point is.

dgutov:
Speaking of `inline-popup`, could you take a look at https://github.com/cpitclaudel/quick-peek?

Could we simply reuse that package, or is it missing some feature, like being able to draw our own borders?

dgutov:
> Don't you think that both the posframe popup and the inline one should be aligned vertically with the indicator on the fringe?

...and I think it would be nice if the "inline popup" variant entirely hid the hunk it was displaying. Like, temporarily replaced the buffer contents covered by the hunk overlay.

alvarogonzalezsotillo:
I will get back to this in a few days, I didn't bring my laptop with me this holidays.

> Speaking of inline-popup, could you take a look at https://github.com/cpitclaudel/quick-peek?

i didn't know about this package. It seems it has not built-in scroll, but if we hide the current hunk and overlap the diffs, then it is better to use the buffer scroll.

I will implement a backend using this package as  soon as I come back to office.


> ...and I think it would be nice if the "inline popup" variant entirely hid the hunk it was displaying.

I supose the hunk can be "folded", like `origami` package. I will investigate the `origami` source code to know the way it hides regions of the buffer.

> Don't you think that both the posframe popup and the inline one should be aligned vertically with the indicator on the fringe?

For big hunks (bigger that the current window), it would be necesary to scroll the buffer to the initial line of the hunk. I think it is better to not change the vertical scroll of the buffer, but we can implement a customizable flag to control this behaviour.

The contents of the posframe are currently scrolled to the clicked line. If  the posframe is positioned at the beggining of the hunk, maybe the contents of the posframe should not be initially scrolled.

dgutov:
> I will get back to this in a few days, I didn't bring my laptop with me this holidays.

Please take your time.

> i didn't know about this package. It seems it has not built-in scroll, but if we hide the current hunk and overlap the diffs, then it is better to use the buffer scroll.

To scroll the contents inside the popup? Or have the popup sometimes span beyond the current window? Either way, perhaps extending it would take less code that maintaining a separate implementation.

I'm not saying we have to use it, mind you, just that reusing it might same us some trouble with packaging. And Clement is a pretty friendly guy, so if there's only some minor omissions, we can file a feature request or two. Or a patch.

https://github.com/cpitclaudel/quick-peek/issues/12 seems relevant, though.

> I supose the hunk can be "folded", like origami package. I will investigate the origami source code to know the way it hides regions of the buffer.

One way to do that is to put the `invisible` property on the overlay that covers it. That might have some weird interactions with the inline-popup (try and see, maybe).

> For big hunks (bigger that the current window), it would be necesary to scroll the buffer to the initial line of the hunk. I think it is better to not change the vertical scroll of the buffer, but we can implement a customizable flag to control this behaviour.

I'm not sure scrolling the buffer to the beginning of the hunk will be a problem. But if you want an option, sure.

> The contents of the posframe are currently scrolled to the clicked line. If the posframe is positioned at the beggining of the hunk, maybe the contents of the posframe should not be initially scrolled.

Maybe.

There is a more pressing problem with the posframe, though: if the hunk is too long, the popup is moved vertically to fit on the screen. Perhaps if we should limit its height, to lower the odds of this happening.

alvarogonzalezsotillo:
I have tested `quick-peek` and it is not possible to scroll through a big hunk, since the buffer 'jumps' from the beggining of the overlay to the end.

> There's also https://github.com/jpablobr/scrollable-quick-peek, published just recently.


Commit 6651880 is an initial implementation of a `scrollable-quick-peek` backend. It lacks some functionality (next and previous hunk or text colors, for example), but it is almost copy-and-paste from other implementations.

There are some limitations of `scrollable-quick-peek` that can be solved via a pull request to that project:
- There is no visual indication of beggining/end  of the hunk
- It is not currently possible to add some text in the header/footer of the quick-peek
- In my personal opinion, my `inline-popup` is nicer without further customization 😅

> But we probably won't be able to depend on it, for the lack of copyright assignment it won't be in ELPA.

I think we can continue with `inline-popup` to avoid external dependencies. We can always switch to `scrollable-quick-peek` if it becomes available on ELPA. The change should be almost trivial.

alvarogonzalezsotillo:
> ...and I think it would be nice if the "inline popup" variant entirely hid the hunk it was displaying. Like, temporarily replaced the buffer contents covered by the hunk overlay.

Done in 119f530

> Right now, the posframe popup especially looks kinda random, showing up wherever the point is.

If `diff-hl-show-hunk-ensure-visible` is `t`(default), the point is moved to the hunk start, so I think this issue is solved.

alvarogonzalezsotillo:
Since `diff-hl-show-hunk-ensure-visible` moves the buffer point to the beggining of the hunk, I think it has no sense to highlight the line where the popup was invoked.

I have added `diff-hl-show-hunk-highlight-clicked-line` custom variable, to control if the line should be highlighted or not. Also, I made the `line` parameter of the backend function optional.

Currently, I think `ensure-visible` and `highlight-clicked-line` are a little bit messy, but I don't know how to make them clearer:

- `ensure-visible` controls two behaviours: (1) It scrolls the buffer to show the entire hunk and (2) move the point to the beggining of the hunk. Two variables, maybe?

- If `ensure-visible` is `t` then `highlight-clicked-line` should be `nil`. Only one variable, maybe?

dgutov:
Thanks for the updates, I have tried the latest version now.

I see what you're saying about `quick-peek`. Since `inline-popup` does look better already, there is no reason to choose the other version. I was just thinking of a way to avoid bundling some extra code, but that's a lesser problem than having subpar UI. So let's not include `scrollable-quick-peek` based backend now.

> If diff-hl-show-hunk-ensure-visible is t(default), the point is moved to the hunk start, so I think this issue is solved

Just to be clear, do you want to keep the option (and the possibility of both behaviors) because you personally the non-default one? Or because it's really easy to implement?

Generally, I avoid having options unless explicitly requested by someone because they are a source of bugs and an extra maintenance burden.

> I have added diff-hl-show-hunk-highlight-clicked-line custom variable, to control if the line should be highlighted or not. 

I'm not sure about it. Do you find that behavior helpful? Personally, I think about this feature in terms of hunks, and the current line only determines which hunk is current. So one solution would be to just remove the variable.

Here's one new bug which I'm not sure how hard would be to fix: with the `inline-popup` backend, if you press `n` followed by `p`, the popup will continue showing the same hunk (and point will seemingly jump from the position just after the overlay to visual end of the line preceding it). If it's easy enough to fix, let's go with the current behavior, hiding the current hunk's region, but otherwise, or maybe as some ideas for future development, here's how some other editors behave here:

* VS Code's popup looks quite similar to `inline-popup`, only with smaller buttons, I guess. It's always shown right after the end of the current hunk. If the hunk is very long, the window is scrolled to accommodate the popup, that that's all. And that seems kind of natural, so keeping the hunk visible would be all right with me with this approach.

* IDEA (last year's Community Edition) has a fiddlier behavior. Might be even a little too clever. It does show the popup right below the line you clicked on the gutter. But then, in the gutter, it makes sure not to show too much. If the hunk was addition-only, the popup doesn't show any text, only the buttons (up/down/show diff/revert). If the hunk was removal-only, it shows the removed text (the indicator is tiny in that case, so there's no question where to show the popup). If the hunk was a mixed change, then it shows the previous text in the hunk, optionally with the "refined diff" effect which we know by `diff-mode`. This might be a bit harder to implement, but the ideas seem sound. Especially the one when we don't show the addition-only hunk's content in the popup.

One possible way to combine these two would be to always show the popup below the end of the current hunk, and in the popup, show the "previous" contents of the lines covered by the hunk. `diff-mode` knows how to add syntax highlighting, so we can implement that too (perhaps later).

Anyway, it's just an idea. What I wanted to say is I'm not married to the exact changes in behavior I asked for; they just made sense at the time, but you can pick the behavior you like first (or one that is easier to implement and make work reliably), and we can get that in, and add some options after in case we differ in preferences.

Another couple of bugs I found:

* Inline-popup always seems to show 2 empty lines at the end inside. The last line may be unavoidable, but it would be great to remove at least one of them.
* If I move around with posframe's popup (`n` or `p`, once or a few times) after I dismiss it, it doesn't return the focus to the main Emacs window. This might be a bug in `posframe`, though.

dgutov:
Here are some gifs showing the workflows in those two editors:

https://user-images.githubusercontent.com/22350/31500262-43db52de-af67-11e7-9393-bddefaded502.gif
https://cloud.githubusercontent.com/assets/1312873/17848825/86bd607c-684d-11e6-86d6-f8964bd422f4.gif

Hope they can serve as an illustration.

alvarogonzalezsotillo:
> Inline-popup always seems to show 2 empty lines at the end inside. The last line may be unavoidable, but it would be great to remove at least one of them.

I removed one of the empty lines. The other one is used to draw the horizontal line over the last line.

> If I move around with posframe's popup (n or p, once or a few times) after I dismiss it, it doesn't return the focus to the main Emacs window. This might be a bug in posframe, though.

I only reproduce this bug in Gnome (not in KDE). I have added an extra `select-frame-set-input-focus` in a timer and seems to work.

> Generally, I avoid having options unless explicitly requested by someone because they are a source of bugs and an extra maintenance burden.

I agree. I have removed `diff-hl-show-hunk-ensure-visible` . I think I will remove also `diff-hl-show-hunk-highlight-clicked-line`

alvarogonzalezsotillo:
> One possible way to combine these two would be to always show the popup below the end of the current hunk, and in the popup, show the "previous" contents of the lines covered by the hunk. 

The commit 081e202 tries to implement this feature for `inline-popup`. If the variable (not so originaly named) `diff-hl-show-hunk-inline-popup-smart-lines` is `t`, the inline popup only shows the lines starting with `-`. If the hunk consist only on additions, then they are showed, because an empty content in the popup seems strange to me.

There is a bug I'm working on: when going to previous hunk from the first hunk, `No further hunks found` appears in minibuffer instead of `There is no previous change`.

dgutov:
> I removed one of the empty lines. The other one is used to draw the horizontal line over the last line.

You can use the `:overline` property instead. And I'd also like to see no empty line above the header. Like this:

```diff
diff --git a/inline-popup.el b/inline-popup.el
index 1de3826..ff554ce 100644
--- a/inline-popup.el
+++ b/inline-popup.el
@@ -87,7 +87,7 @@ Default for CONTENT-SIZE is the size of the current lines"
          (header (if (< new-width 0) "" header))
          (new-width (- width (length header) (length scroll-indicator)))
          (line (propertize (concat (inline-popup--separator new-width) header scroll-indicator ) 'face '(:underline t))))
-    (concat "\n" line "\n") ))
+    (concat line "\n") ))
 
 (defun inline-popup--compute-footer (width &optional footer)
   "Compute the header of the popup, with some WIDTH, and some optional FOOTER text."
@@ -96,9 +96,9 @@ Default for CONTENT-SIZE is the size of the current lines"
          (new-width (- width (length footer) (length scroll-indicator)))
          (footer (if (< new-width 0) "" footer))
          (new-width (- width (length footer) (length scroll-indicator)))
-         (blank-line (propertize (inline-popup--separator width) 'face '(:underline t)))
-         (line (propertize (concat (inline-popup--separator new-width) footer scroll-indicator))))
-    (concat "\n" blank-line "\n" line)))
+         (line (propertize (concat (inline-popup--separator new-width) footer scroll-indicator)
+                           'face '(:overline t))))
+    (concat "\n" line)))
 
 (defun inline-popup--separator (width &optional sep)
   "Return the horizontal separator with character SEP and a WIDTH."
```

> I only reproduce this bug in Gnome (not in KDE). I have added an extra select-frame-set-input-focus in a timer and seems to work.

Thanks. Yeah, GNOME is the trickiest with child frames, requiring more workarounds than the most. I wonder of other packages using posframe has seen similar bugs, and how they solved it.

> I agree. I have removed diff-hl-show-hunk-ensure-visible . I think I will remove also diff-hl-show-hunk-highlight-clicked-line

Thanks.

> The commit 081e202 tries to implement this feature for inline-popup. If the variable (not so originaly named) diff-hl-show-hunk-inline-popup-smart-lines is t, the inline popup only shows the lines starting with -.

Thank you. And when it's non-nil, I think it would be natural to have `diff-hl-show-hunk-inline-popup-hide-hunk` set to `nil`. That way, the user sees the current buffer contents, and the previous ones, but no duplication.

> If the hunk consist only on additions, then they are showed, because an empty content in the popup seems strange to me.

I think, with `diff-hl-show-hunk-inline-popup-hide-hunk`=`nil`, showing the pure-additions hunk seems unnecessary. You're right about plain empty popup being a problem, though. Maybe it should show something, maybe it should display some clarification text, or maybe it should ignore/skip such hunks because it doesn't have anything to show for them that's not already on the screen. Anyway, that can be a feature update for later.

Will you remove the `scrollable-quick-peek` backend?

alvarogonzalezsotillo:
> There is a bug I'm working on: when going to previous hunk from the first hunk, No further hunks found appears in minibuffer instead of There is no previous change.

Solved.

> You can use the :overline property instead. And I'd also like to see no empty line above the header. 

There is a problem in console (not GUI): console doesn't not support overline, so I simulated it with an empty line with underline. I added a `display-graphic-p` test, hope it will have no problem with the GUI and the console is used at the same time (via `emacsclient`)

> And when it's non-nil, I think it would be natural to have diff-hl-show-hunk-inline-popup-hide-hunk set to nil. 

Currently, I prefer `-inline-popup-hide-hunk` set to `t` and `-smart-lines` set to `t` :) For me is like a wayback machine. I think I have set the default values to your preferences, tell me if not.

> You're right about plain empty popup being a problem, though. Maybe it should show something, maybe it should display some clarification text, 

I have added the `diff-hl-show-hunk--no-lines-removed-message` constant, that is shown when `-smart-lines` set to `t` and the hunk has only additions.

> Will you remove the scrollable-quick-peek backend?

I forgot it. Removed.

dgutov:
> Solved.

Thanks.

> There is a problem in console (not GUI): console doesn't not support overline, so I simulated it with an empty line with underline. I added a display-graphic-p test, hope it will have no problem with the GUI and the console is used at the same time (via emacsclient)

I wonder if console users would prefer no extra line, even if the horizontal line can't be shown. Anyway, we can leave that to future bug reports.

But note that you missed the first bit from the diff I posted: the one that removed a newline from in front of the header as well (in `inline-popup--compute-header`).

> Currently, I prefer -inline-popup-hide-hunk set to t and -smart-lines set to t :) For me is like a wayback machine. I think I have set the default values to your preferences, tell me if not.

All right, so it seems we should keep the both option variables. But is it working right for you?

If you prefer for them to be both `t`, perhaps this piece of code should do something else?

```el
+         (smart-lines (and diff-hl-show-hunk-inline-popup-smart-lines (not diff-hl-show-hunk-inline-popup-hide-hunk)))
```

> I have added the diff-hl-show-hunk--no-lines-removed-message constant, that is shown when -smart-lines set to t and the hunk has only additions.

Perhaps it should say "no lines removed".

Something else I've been thinking: since `diff-hl-show-hunk` is going to be a built-in feature, there's no need for it to require a minor mode to be accessible. It can just be in `diff-hl-mode` by default. The mouse-mode can still be a move, or it can be determined by a user option. WDYT?

alvarogonzalezsotillo:
> But note that you missed the first bit from the diff I posted: the one that removed a newline from in front of the header as well (in inline-popup--compute-header).

You are right, removed that extra new line.

> All right, so it seems we should keep the both option variables. But is it working right for you? If you prefer for them to be both t, perhaps this piece of code should do something else?

I admit I have changed my mind a couple of times:

|                 | hide-hunk                   | not hide-hunk                        |
|-----------------|-----------------------------|--------------------------------------|
| smart-lines     | (1) Sort of wayback machine | (2) Your preferences (default)       |
| not smart-lines | (3) My preferences          | (4) New lines are shown twice, noisy |

I have tried to avoid (4), and convert it to (3), but the logic was wrong. I have changed an `and` for an `or` to avoid (4), and convert it to (2)

alvarogonzalezsotillo:
> It can just be in diff-hl-mode by default. The mouse-mode can still be a minor mode, or it can be determined by a user option. WDYT?

I think mouse-mode and global-minor-mode should be a separate minor mode, to avoid conflicts with other packages, but the rest of key bindings can be in `diff-hl-mode`

alvarogonzalezsotillo:
> Something else I've been thinking: since diff-hl-show-hunk is going to be a built-in feature, there's no need for it to require a minor mode to be accessible. It can just be in diff-hl-mode by default. 

I'm about to implement this comment. So, `diff-hl.el` should require `diff-hl-show-hunk.el`?

dgutov:
> So, diff-hl.el should require diff-hl-show-hunk.el?

Not necessary: you just add `;;;###autoload` above the command's definition and add its name to `diff-hl-mode`'s keymap. Then whenever somebody uses it, the package will be autoloaded.

dgutov:
> I have tried to avoid (4), and convert it to (3), but the logic was wrong. I have changed an and for an or to avoid (4), and convert it to (2)

Thanks for the table, it confirms that we do need both variables. But I think we'll just remove the condition for `smart-lines`: while it makes (4) impossible, let's just make it the user's choice. Doing a runtime check like this would be useful if we also predicated it on something that changes between runs/hunks/files or so on, but this looks like a decision the user can make on their own.

Finally, it's high time to start merging this. I'll try to make time this weekend to merge the "meat" of this PR (the modes, commands, and the keymap changes), but simplify the README change for now.

Redoing the README will come after (and BTW, I'll need the name of the tool you used for the gifs ;).

alvarogonzalezsotillo:
> But I think we'll just remove the condition for smart-lines: while it makes (4) impossible, let's just make it the user's choice.

Done

> I'll need the name of the tool you used for the gifs ;)

If you mean the displayed keystrokes, I use [screenkey](https://gitlab.com/screenkey/screenkey.git). The version in Ubuntu repositories is not very up to date, so I use the repository version (just checkout and run `./screenkey --opacity 0.0`) . I think [it only works on X11](https://www.reddit.com/r/linuxquestions/comments/kahs26/screenkey_for_wayland/) :(

If you mean the screencast, I use [OBS](https://obsproject.com/) and then convert the video to gif with 

   ```
     # https://superuser.com/questions/556029/how-do-i-convert-a-video-to-gif-using-ffmpeg-with-reasonable-quality
     INPUT="$1" 
     OUTPUT="$2"
     WIDTH="$3"
     ffmpeg  -i "$INPUT" -vf "fps=10,scale=$WIDTH:-1:flags=lanczos,split[s0][s1];[s0]palettegen[p];[s1][p]paletteuse" -loop 0 "$OUTPUT"
   ```

dgutov:
Excellent, thank you.

dgutov:
@purcell Hi Steve! Would you say it's okay for `diff-hl` to "incubate" the `inline-popup` package inside it with its current file name, or should it be renamed (together with all symbols in it) to match the name `diff-hl`?

purcell:
> Would you say it's okay for `diff-hl` to "incubate" the `inline-popup` package inside it with its current file name, or should it be renamed (together with all symbols in it) to match the name `diff-hl`?

It should match the name prefix, or we could have the two packages coexist in the same repo, ie. 2 recipes. That's a bit fiddly though. I'd suggest incubating with the `diff-hl-` prefix, then later split it out and drop the prefix.

dgutov:
> I'd suggest incubating with the diff-hl- prefix, then later split it out and drop the prefix.

Thanks for the reply, guess I'll go that route indeed.

alvarogonzalezsotillo:
I didn't notice the new `diff-hl-amend-mode` of the last month.

The header of the inline popup should be updated with the revision the diff is computed against, something like:

```
   -"Diff with HEAD"
   +(concat "Diff with " (if diff-hl-reference-revision diff-hl-reference-revision "last revision"))
```

Since you are merging this PR, do you prefer to make this change yourself? Or is it better if I merge `master` in this PR and push this change?

dgutov:
@alvarogonzalezsotillo Please go ahead.

alvarogonzalezsotillo:
@dgutov Just in case, you can find in https://github.com/alvarogonzalezsotillo/diff-hl/tree/unincubating-inline-popup a branch that has merged master, this PR and `inline-popup`-to-`diff-hl-inline-popup` refactor.

dgutov:
Thanks! I've done the merge now, and simplified some bits and pieces.

Please take a look, see whether everything still works to your expectations, and let's continue fixing and improving the feature in new issues and PRs.

I hope to get around to refreshing the images (and adding a gif or two) sometime soon.

PR #150, opened 2020-11-17
Add global-diff-hl-amend-mode to diff-hl custom group

tarsius:
Emacs 28's byte-compiler was complaining about it not
being a member of any group.


dgutov:
Thanks!

PR #151, opened 2020-11-27
New advice to diff-hl-diff-show-hunk to work in diff-hl-flyspell-mode

alvarogonzalezsotillo:
Inspired in https://github.com/dgutov/diff-hl/pull/147#issuecomment-733511417

This is a new advice to `diff-hl-diff-show-hunk` to work in `diff-hl-flyspell-mode`.

This function is used by `diff-hl-show-hunk` (https://github.com/dgutov/diff-hl/pull/147), so `diff-hl-show-hunk` doesn't need to save the buffer.


alvarogonzalezsotillo:
See https://github.com/dgutov/diff-hl/pull/147#issuecomment-735318376

PR #156, opened 2021-02-20
Use vc-read-revision in some cases set-reference-rev can't deduce the revision

alvarogonzalezsotillo:
I think the revision deduction process of `diff-hl-set-reference-rev` can be improved if `vc-read-revision` is used when in a vc-backed buffer.

It seems to work with and without [helm](https://emacs-helm.github.io/helm/). I have not tested it with other completion frameworks.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -725,7 +725,7 @@ The value of this variable is a mode line template as in
                                             nil
                                             at-point
                                             nil))
-                        (thing-at-point 'symbol t))))
+                        at-point)))
     (when (string= "" revision)
       (user-error "Empty revision"))
     revision))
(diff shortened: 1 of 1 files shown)


dgutov:
Hi! Thanks for the report.

I've made a similar change, but one that calls `vc-read-revision` in any case, using one of the deduced values as the default.

Let me know how you like it, and whether the extra RET tends to be too much.

/Cc @AmaiKinono

alvarogonzalezsotillo:
I tried to mimic the previous behavior, asking for a revision if not sure.

But I prefer this solution, because there is an option to confirm the revision before setting it.

AmaiKinono:
I like it ;) Seems for git repository, I can pick a branch name as the reference rev this way.

PR #158, opened 2021-02-27
remove unused `buffer` argument in diff-hl-make-temp-file-name

detvdl:
`buffer` argument of `diff-hl-make-temp-file-name` was unused.

Emacs byte compiler started complaining about an unused lexical argument, perhaps due to the [new-ish byte-compiler check for missing dynamic variable declarations](https://github.com/emacs-mirror/emacs/blob/3a7702135377402015368c8307503b62c421655c/etc/NEWS#L2620)??

Occurred on:

Emacs 28.1 compiled from source origin/master branch on 27-02-2021 on Ubuntu 18.04

Configured with `--with-mailutils --with-imagemagick --with-xwidgets`, all the rest is default.

`GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.30, cairo version 1.15.10)
of 2021-02-27`

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -729,7 +729,7 @@ The value of this variable is a mode line template as in
       (add-hook 'vc-checkin-hook 'diff-hl-dir-update t t)
     (remove-hook 'vc-checkin-hook 'diff-hl-dir-update t)))
 
-(defun diff-hl-make-temp-file-name (buffer rev &optional manual)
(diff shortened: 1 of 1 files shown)


dgutov:
Hi! Thank you for the reminder.

It's a leftover from the recent feature merge (previously it was in a file without lexical-binding enabled). I just put off dealing with it because it requires some docstring updates too with consideration of the semantics.

PR #160, opened 2021-03-18
Last line of inline-popup properly underlined in terminal.

alvarogonzalezsotillo:
This minor change fixes the last line of the inline popup in terminal. The last content line had an extra underlined character at the end.

Also, `diff-hl-show-hunk--goto-hunk-overlay` is now more tolerant to previous errors.


dgutov:
Thanks!

PR #162, opened 2021-04-21
Fix building libraries using make

tarsius:
- Compile all libraries, including the main one.
- Make the "all" make target depend on "compile".
- Add the current directory to the load-path,
  so that compiling doesn't fail completely.


dgutov:
Thanks, Jonas.

PR #163, opened 2021-05-18
Silence byte-compiler

tarsius:
The byte-compiler has a second complaint, which this pr does *not* tackle:

```
In diff-hl-changes-from-buffer:
lib/diff-hl/diff-hl.el:332:12: Warning: ‘diff-auto-refine-mode’ is an obsolete
    function (as of 27.1); set ‘diff-refine’ instead.
```

This is weird because that is used as a *variable*, not a function.  Sure, the variable is obsolete too, but the byte-compiler seems confused.  Or can you see what's going on?

Anyway, this could be suppressed by wrapping the `let[*]` with `with-no-warnings`.  But maybe the scope of that could be reduced to avoid accidentally suppressing other legitimate warnings in the future?

Could you please take care of this please? I regularly recompile all the packages I have installed and try to fix all the new warnings, but this is a hairy case that I think the maintainer should investigate instead.

Thanks!

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -885,6 +885,8 @@ the `diff-program' to be in your `exec-path'."
               (t (memq major-mode diff-hl-global-modes)))
     (turn-on-diff-hl-mode)))
 
+(declare-function vc-annotate-extract-revision-at-line "vc-annotate")
+
 ;;;###autoload
 (defun diff-hl-set-reference-rev (rev)
   "Set the reference revision globally to REV.


dgutov:
Thanks for the patch.

> Sure, the variable is obsolete too, but the byte-compiler seems confused. Or can you see what's going on?

Probably a bug in the byte-compiler, but even after it's fixed, the warning will still be there, right? Just about a variable.

> But maybe the scope of that could be reduced to avoid accidentally suppressing other legitimate warnings in the future?

The concern is valid, and I'm not sure how to reduce the scope in a let-binding (and there are 3 of those).

My general policy has been to ignore obsoletion warnings when there is no solution that would work in the oldest supported version of Emacs, which is 25.1 in this case. `with-no-warnings` would be an exception, but it has downsides, as you say.

Suggestions welcome, of course.

dgutov:
There is also the macro `with-suppressed-warnings` now, but it was added in 27.1 as well.

tarsius:
> Thanks for the patch.
> 
> > Sure, the variable is obsolete too, but the byte-compiler seems confused. Or can you see what's going on?
> 
> Probably a bug in the byte-compiler, but even after it's fixed, the warning will still be there, right? Just about a variable.

Yes.

> > But maybe the scope of that could be reduced to avoid accidentally suppressing other legitimate warnings in the future?
> 
> The concern is valid, and I'm not sure how to reduce the scope in a let-binding (and there are 3 of those).

What I meant is this: I don't know for whose benefit you are let-binding `diff-auto-refine-mode` to `nil`. If it is only for, say, `diff-beginning-of-hunk`, then you could relocate the `let` so that it only wraps around that. If you then wrap the `with-no-warnings` around just that in turn, then the only future warnings that we might potentially miss, would all concern just `diff-beginning-of-hunk`.

> My general policy has been to ignore obsoletion warnings when there is no solution that would work in the oldest supported version of Emacs, which is 25.1 in this case. `with-no-warnings` would be an exception, but it has downsides, as you say.

Well... my policy has been more OCD-oriented: no package I have installed shall emit compile warnings. :grinning:

(With obvious exceptions like `undo-tree`, because getting that fixed is hopeless.)

> Suggestions welcome, of course.

"Wrap `with-no-warnings` around as little code as possible" is all I got right now.

dgutov:
> Well... my policy has been more OCD-oriented: no package I have installed shall emit compile warnings. grinning

I suspect you might have been unhappy with the packages I maintain on multiple occasions on that point. :-)

People do ask about the warnings, though, and/or provide minor fixes. That helps.

> "Wrap with-no-warnings around as little code as possible" is all I got right now

And this is done now :ok_hand:

tarsius:
I might have addressed or fixed such issues in your libraries once or twice. :grin: 

Thanks!

dgutov:
Let this be known as the day I've added the "byte-compie warnings" step in the pipeline here :zap:

PR #165, opened 2021-06-30
bury show hunk buffers on quit

wyuenho:
Currently, after quitting the show-hunk overlay/posframe, the show hunk buffers are at the top of the buffer list, which breaks the usual expectation when switching buffers. Ancillary buffers should not ever be at the top of the buffer list after quitting their modes.


dgutov:
Thanks!

PR #167, opened 2021-07-15
makes sure a buffer `diff-hl-show-hunk-diff-buffer-name` exists...

fgeller:
before burying it. Not sure if this is the idiomatic solution for diff-hl, could maybe group it into a loop or maybe you're looking for an entirely different solution. either way, thanks for a cool mode and maybe this helps!

Fixes this error:

```
Debugger entered--Lisp error: (error "No such live buffer *diff-hl-show-hunk-diff-buffer...")
  signal(error ("No such live buffer *diff-hl-show-hunk-diff-buffer..."))
  error("No such live buffer %s" "*diff-hl-show-hunk-diff-buffer*")
  window-normalize-buffer("*diff-hl-show-hunk-diff-buffer*")
  bury-buffer("*diff-hl-show-hunk-diff-buffer*")
  diff-hl-show-hunk-hide()
  diff-hl-show-hunk()
  funcall-interactively(diff-hl-show-hunk)
  call-interactively(diff-hl-show-hunk record nil)
  command-execute(diff-hl-show-hunk record)
  execute-extended-command(nil "diff-hl-show-hunk" nil)
  funcall-interactively(execute-extended-command nil "diff-hl-show-hunk" nil)
  call-interactively(execute-extended-command nil nil)
  command-execute(execute-extended-command)
```

### DIFF
--- a/diff-hl-show-hunk.el
+++ b/diff-hl-show-hunk.el
@@ -108,6 +108,9 @@ corresponding to the clicked line in the original buffer."
   (with-current-buffer (get-buffer-create diff-hl-show-hunk-buffer-name)
     (read-only-mode -1)
     (erase-buffer))
+  (with-current-buffer (get-buffer-create diff-hl-show-hunk-diff-buffer-name)
+    (read-only-mode -1)
+    (erase-buffer))
   (bury-buffer diff-hl-show-hunk-buffer-name)
   (bury-buffer diff-hl-show-hunk-diff-buffer-name)
   (when diff-hl-show-hunk--hide-function


dgutov:
I'm curious how @wyuenho never noticed/triggered this.

dgutov:
@fgeller Thanks for the patch/report.

I've just pushed a slightly shorter fix to master. Could you try it first?

wyuenho:
Probably because I've never tested #165 where that buffer didn't exist 😅

wyuenho:
This PR works well.

dgutov:
> Probably because I've never tested #165 where that buffer didn't exist sweat_smile

It always doesn't exist at the beginning of an Emacs session, does it? The first time one calls `diff-hl-show-hunk`.

> This PR works well.

Could you try the fix I already pushed to master?

wyuenho:
> It always doesn't exist at the beginning of an Emacs session, does it? The first time one calls diff-hl-show-hunk.

I've never tried #165 on a fresh session.

Master works well too.

dgutov:
Thanks for checking.

PR #172, opened 2021-10-08
Fix #171 make Magit hooks great again

### LINKED ISSUE #171: diff-hl-magit-post-refresh doesn't work anymore
I'm on the emacs-28 branch HEAD and using latest Magit and libgit from Melpa, I've followed the README to do

```elisp
(with-eval-after-load 'magit
    (add-hook 'magit-post-refresh-hook 'diff-hl-magit-post-refresh))
```

However, the post-fresh hook doesn't seem to be able to call `diff-hl-update` update anymore after `magit-commit`. The reason seems to be after `magit-commit`, the file is considered as `up-to-date` by git, so neither of the branches will hit `diff-hl-update`.

What I don't understand is, why is `diff-hl-magit-post-refresh` so complicated? Why not just call `diff-hl-update` as long as the file still exists on disk and in vc?

wyuenho:
Re: #171 calling `magit-unstaged-files` when magit is refreshing will always result in nil because magit only refreshes when everything is done.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -861,47 +861,36 @@ The value of this variable is a mode line template as in
 (declare-function magit-toplevel "magit-git")
 (declare-function magit-unstaged-files "magit-git")
 
-(defvar diff-hl--magit-unstaged-files nil)
+(defvar diff-hl--vc-states nil)
 
 (defun diff-hl-magit-pre-refresh ()
   (unless (and diff-hl-disable-on-remote
                (file-remote-p default-directory))
-   (setq diff-hl--magit-unstaged-files (magit-unstaged-files t))))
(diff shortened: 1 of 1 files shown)


dgutov:
I think the most productive approach here would be to ping Jonas from time to time, until he either provides his own solution, or blesses one of yours. I don't have the mental bandwidth right now to get into which of the approaches is better for Magit (which I don't use).

Thanks either way.

tarsius:
Yeah, this has kept falling through the cracks. But it is a "just one more thing" issue, and I am currently trying to take a break from those (see https://www.reddit.com/r/emacs/comments/11f7urf/thanks_for_your_support_and_patience/). I still plan to look into this again at some point, but won't get around to that for at least another few weeks.

PR #178, opened 2022-01-05
Mention that VC must be enabled

raxod502:
Some people disable VC because it fixes various oddities in Emacs and the same features are provided by other packages such as Magit. In this case, diff-hl will silently do nothing, which is confusing. This PR adds a note to the README clarifying the requirements for using diff-hl.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,9 @@ Put this into your init script:
 (global-diff-hl-mode)
 ```
 
+You must also ensure that VC is enabled (e.g. `vc-handled-backends` is
+not nil).
+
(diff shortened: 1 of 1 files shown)


dgutov:
Sure, thanks.

dgutov:
Which oddities are you referring to, though?

I had `vc-handled-backends` set to nil in the past at some point, but that was solely to improve the performance.

raxod502:
I disabled it because I was not able at the time to find any other way to remove the message `Followed link to /some/file` every time you open a file through a symlink, per https://stackoverflow.com/a/45473238.

PR #179, opened 2022-01-17
Add new option diff-hl-disable-on-remote

Ergus:
To inhibit diff-hl in remote buffers. This option is useful with slow
tramp connections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,3 +127,12 @@ When using Magit 2.4 or newer, add this to your init script:
 (add-hook 'magit-pre-refresh-hook 'diff-hl-magit-pre-refresh)
 (add-hook 'magit-post-refresh-hook 'diff-hl-magit-post-refresh)
 ```
+
+Tramp
+-----
+
+`diff-hl` should just work with Tramp. But sometimes a performance
+penalty may be experienced in slow or high latency connections. If
+you experiment such issue then an option may be to set
+`diff-hl-disable-on-remote` to `t` this will inhibit some `diff-hl`
+actions and hooks for the remote buffers.
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -106,6 +106,11 @@
   :group 'diff-hl
   :type 'boolean)
 
+(defcustom diff-hl-disable-on-remote nil
+  "Non-nil will disable `diff-hl' on remote buffers."
+  :group 'diff-hl
+  :type 'boolean)
+
 (defcustom diff-hl-ask-before-revert-hunk t
   "Non-nil to ask for confirmation before revert a hunk."
   :group 'diff-hl
@@ -851,10 +856,14 @@ The value of this variable is a mode line template as in
 (defvar diff-hl--magit-unstaged-files nil)
 
 (defun diff-hl-magit-pre-refresh ()
-  (setq diff-hl--magit-unstaged-files (magit-unstaged-files t)))
+  (unless (and diff-hl-disable-on-remote
+               (file-remote-p default-directory))
+   (setq diff-hl--magit-unstaged-files (magit-unstaged-files t))))
 
 (defun diff-hl-magit-post-refresh ()
-  (let* ((topdir (magit-toplevel))
+  (unless (and diff-hl-disable-on-remote
+               (file-remote-p default-directory))
+   (let* ((topdir (magit-toplevel))
          (modified-files
           (mapcar (lambda (file) (expand-file-name file topdir))
                   (delete-consecutive-dups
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -127,3 +127,12 @@ When using Magit 2.4 or newer, add this to your init script:
 (add-hook 'magit-pre-refresh-hook 'diff-hl-magit-pre-refresh)
 (add-hook 'magit-post-refresh-hook 'diff-hl-magit-post-refresh)
 ```
+
+Tramp
+-----
+
+`diff-hl` should just work with tramp. But sometimes a performance
```
dgutov, line 134: Please capitalize "Tramp".

README.md
```diff
@@ -127,3 +127,12 @@ When using Magit 2.4 or newer, add this to your init script:
 (add-hook 'magit-pre-refresh-hook 'diff-hl-magit-pre-refresh)
 (add-hook 'magit-post-refresh-hook 'diff-hl-magit-post-refresh)
 ```
+
+Tramp
+-----
+
+`diff-hl` should just work with tramp. But sometimes a performance
+penalty may be experimented in slow or high latency connections. If
```
dgutov, line 135: experimented -> experienced

And on the next line again.

README.md
```diff
@@ -127,3 +127,12 @@ When using Magit 2.4 or newer, add this to your init script:
...
 (add-hook 'magit-post-refresh-hook 'diff-hl-magit-post-refresh)
 ```
+
+Tramp
+-----
+
+`diff-hl` should just work with tramp. But sometimes a performance
+penalty may be experimented in slow or high latency connections. If
+you experiment such issue then an option may be to set
+`diff-hl-disable-on-remote` to `t` this will inhibit some `diff-hl`
```
dgutov, line 137: It will disable (almost?) all the features, would it not?

diff-hl.el
```diff
@@ -851,10 +856,14 @@ The value of this variable is a mode line template as in
 (defvar diff-hl--magit-unstaged-files nil)
 
 (defun diff-hl-magit-pre-refresh ()
-  (setq diff-hl--magit-unstaged-files (magit-unstaged-files t)))
+  (unless (and diff-hl-disable-on-remote
+	       (file-remote-p default-directory))
```
dgutov, line 860: It seems like you're using tabs for indentation here. Could you untabify all the added lines?

dgutov:
Looking good otherwise.

README.md
Ergus: `diff-hl-dired` and the others maybe??

Ergus:
I think it is fixed now.

README.md
dgutov: `diff-hl-dired` is also disabled, right?

So are there any features that keep working in remote buffers when this variable is non-nil?

Ergus: No, I didn't touch `diff-hl-dired` at all. to disable it on remoted there is `diff-hl-dired-mode-unless-remote`.

I didn't change that because I was not sure how do you prefered to proceed. 
1) Modify diff-hl-dired-mode-unless-remote
2) Create a new function something like diff-hl-dired-mode-turn-on 
```
(defun diff-hl-dired-mode-turn-on ()
  (unless (and diff-hl-disable-on-remote
	       (file-remote-p default-directory))
    (diff-hl-dired-mode)))
```
3) Modify `diff-hl-dired-mode` directly like:

```
(define-minor-mode diff-hl-dired-mode
  "Toggle VC diff highlighting on the side of a Dired window."
  :lighter ""
  (cond
   ((and diff-hl-disable-on-remote
	 (file-remote-p default-directory))
    (diff-hl-dired-mode -1))
   (diff-hl-dired-mode
    (diff-hl-maybe-define-bitmaps)
    (set (make-local-variable 'diff-hl-dired-process-buffer) nil)
    (add-hook 'dired-after-readin-hook 'diff-hl-dired-update nil t))
   (t
    (remove-hook 'dired-after-readin-hook 'diff-hl-dired-update t)
    (diff-hl-dired-clear))))
```
The first option is somehow the simplest, but a bit redundant. The second is almost duplicate the `diff-hl-dired-mode-unless-remote` and the third will limit the user to not add exceptions to this option (something he exceptionally may want to do selectively)

dgutov: Sorry, I mistook `diff-hl-dir-mode` for `diff-hl-dired-mode` inside the definition of `turn-on-diff-hl-mode`. That's where that question came from.

If the behavior in your patch satisfies your requirements, no need to tweak it further. Ultimately my question was about the phrasing in the description, which seems a little too vague. But I can fix it myself, no problem.

dgutov:
Thanks for the PR.

PR #180, opened 2022-01-18
Add .dir-locals.el

yugaego:
I noticed there was a need to untabify in the latest PR, hence should we settle on the code style rules? 
The suggested settings are similar to the ones used by Company, so hopefully, they may work well for diff-hl too.


dgutov:
Thanks!

PR #203, opened 2023-08-07
Silence byte-compiler of Emacs 30

tarsius:


dgutov:
Hi Jonas!

Thanks for the reminder: I deleted the whole top-level form because we only support Emacs 25+ now anyway.

PR #206, opened 2024-01-19
Allow update overlay asynchronously without freezing Emacs

whhone:
Problem: On a slow VC backend, `diff-hl-mode` freezes Emacs because `(diff-hl-changes)` is taking a long time.

Changes:
- Add a custom variable `diff-hl-update-async`
- When true, get the "changes" in a child process, and update the overlay with the result.
- Add a test


whhone:
Closing this pull request because I have a better version of it.

PR #207, opened 2024-01-20
Allow updating indicators asynchronously (without blocking Emacs)

whhone:
Problem: On a slow VC backend, `diff-hl-mode` freezes Emacs because `(diff-hl-changes)` takes a long time.

Changes:
- Add a custom variable `diff-hl-update-with-thread`.
- When set to true, update with a separate thread.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -194,11 +194,12 @@ the NEW revision is not specified (meaning, the diff is against
 the current version of the file)."
   :type 'boolean)
 
-(defcustom diff-hl-update-with-thread nil
+(defcustom diff-hl-update-async nil
   "When non-nil, `diff-hl-update' will run asynchronously.
 
 This can help prevent Emacs from freezing, especially by a slow version
-control (VC) backend."
+control (VC) backend. Remote files will not be affected since this feature
+does not work reliably with them."
   :type 'boolean)
 
 (defvar diff-hl-reference-revision nil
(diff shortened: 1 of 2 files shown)


dgutov:
Hi! That's a neat idea.

Threads are not well-tested and have some known shortcomings (see below), but if this is working well for you, let's try to add this approach.

First of all, this is almost equivalent to `(advice-add 'diff-hl-update :around #'make-thread)`. Although the special handling for remotes makes sense, I've also read something about problems there.

The main thing with threads is that they don't signal errors (one is apparently supposed to call `(thread-last-error)`, but there is no obvious place to do that here). You can observe that by making a typo somewhere inside `diff-hl--update`. The code will stop working, but you won't see any error logged.

So I think we'll need to add some catch-all error handling inside the thread - catch any `error` and at least log a message.

Optionally, I'd look at the cases where the VCS is slow enough that the request might take more time than passes until the next time the user saves the buffer, and try to avoid multiple commands firing in parallel. Some debouncing, maybe. But that can be left as an improvement for later.

Speaking of copyright assignment, you're covered by Google, right?

whhone:
I have updated the commit:
1. catch any error in `diff-hl--update` and log that error, if any
2. add a TODO for debounceing

Thanks for checking the copyright thing. I checked and I think it is okay to contribute to public projects + under GNU GPL 3 + using my company email + no CLA required.

dgutov:
Nice. See more review comments.

diff-hl.el
```diff
@@ -384,38 +391,52 @@ the current version of the file)."
...
+  (if (and diff-hl-update-with-thread
+           ;; Disable threading on the remote file as it is unreliable.
+           (not (file-remote-p default-directory)))
+      ;; TODO: debounce if a thread is already running.
+      (make-thread 'diff-hl--update "diff-hl--update")
+    (diff-hl--update)))
+
+(defun diff-hl--update ()
+  "Updates the diff-hl overlay."
+  (condition-case err
```
dgutov, line 404: How about creating a separate function (e.g. `diff-hl--update-safe`) which would include the `condition-case` form and call this one? Then the change diff should be more compact, and the current execution flow would be preserved more accurately.

dgutov (changes requested):

diff-hl.el
whhone: Done.

diff-hl.el
```diff
@@ -194,6 +194,13 @@ the NEW revision is not specified (meaning, the diff is against
 the current version of the file)."
   :type 'boolean)
 
+(defcustom diff-hl-update-with-thread nil
```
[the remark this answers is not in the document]
whhone: Done.

whhone:
Thanks @dgutov. Comments addressed.

test/diff-hl-test.el
```diff
@@ -126,6 +126,32 @@
     (should (looking-at "added"))
     (should-error (diff-hl-next-hunk) :type 'user-error)))
 
+(diff-hl-deftest diff-hl-indirect-buffer-move-with-thread ()
+  (diff-hl-test-in-source
+    (setq diff-hl-update-with-thread t)
```
dgutov, line 131: Update the name here too.

Also, please set it in `let`, so that it doesn't affect any other loaded tests.

whhone: Thanks for the catch. Fixed.

dgutov:
Thanks!

dgutov:
Heads up: in #246 I have merged an async diffs implementation that does not use a thread, since there arrived some reports of them resulting in some buggy experiences.

To work, the new approach requires a recent Emacs 31 (currently needs to be built from `master`).

One can still switch to using threads which will work with older Emacs releases, with `(setq diff-hl-update-async 'thread)`.

PR #208, opened 2024-01-20
Fix test diff-hl-can-split-away-no-trailing-newline

whhone:
Assert the substring with `cl-search` instead of the whole string with `equal`.


dgutov:
Thanks for reporting the breakage, I guess this repo could use a CI integration.

I prefer to fix it in a different way - still using `equal` because it's the only test that exercises `diff-hl-split-away-changes` so far.

Issue #213, opened 2024-03-22
diff-hl-update-async can break magit-commit


whhone:
Expected behavior: (`COMMIT_EDITMSG` buffer is killed after committing)
1. `magit-commit` uses `with-editor-mode` to edit the commit message (`COMMIT_EDITMSG`). 
2. The user uses `C-c C-c` (`with-editor-finish`) to save the commit message and close the buffer `COMMIT_EDITMSG`.

Actual behavior: (`COMMIT_EDITMSG` buffer is not killed after committing)
1. When `diff-hl-update-async` is enabled and the buffer `COMMIT_EDITMSG` is unsaved,
2. The user uses `C-c C-c` to commit the message, 
3. with-editor will save the buffer, causing `diff-hl-update` to run,
4. the `diff-hl-update-async` thread will prevent the buffer `COMMIT_EDITMSG` to  be killed
5. The next `magit-commit` will try to reuse `COMMIT_EDITMSG`  instead of creating a new buffer.

dgutov:
Hi!

Would it work to just avoid using `diff-hl` in such buffers?

Like this:

```diff
diff --git a/diff-hl.el b/diff-hl.el
index 29dfe31..0d979ed 100644
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -158,7 +158,7 @@ the end position as its only argument."
                  (const :tag "Narrow to the hunk"
                         diff-hl-revert-narrow-to-hunk)))
 
-(defcustom diff-hl-global-modes '(not image-mode)
+(defcustom diff-hl-global-modes '(not image-mode with-editor-mode)
   "Modes for which `diff-hl-mode' is automagically turned on.
 This affects the behavior of `global-diff-hl-mode'.
 If nil, no modes have `diff-hl-mode' automatically turned on.

```

whhone:
`with-editor-mode` is a minor mode and `diff-hl-global-modes` assumes `major-mode`.

But I think this approach works :-)

dgutov:
Okay. So it doesn't use any unique major modes, does it?

You could still add a function to `find-file-hook`, I guess, which would turn off `diff-hl-mode` in certain unsuitable buffers. Can they be determined by file name? We could add regexps to the `diff-hl-global-modes` syntax as well.

Finally, it might be worth looking into the reason for step 4. Are all buffers assigned to live background threads unkillable? Maybe there is a flag somewhere. There is one for processes: `(process-query-on-exit-flag process)`.

[...]

PR #214, opened 2024-03-22
Disable diff-hl-update-async in with-editor-mode

whhone:
It fixes the issue in https://github.com/dgutov/diff-hl/issues/213.


dgutov:
Resolved another way.

PR #225, opened 2024-11-22
Fix #171 by updating changed files in the last commit

### LINKED ISSUE #171: diff-hl-magit-post-refresh doesn't work anymore

wyuenho:
This should finally fix this long standing issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,6 @@ If you're using a version before 2.4.0, it defines `magit-revert-buffer-hook`
 When using Magit 2.4 or newer, add this to your init script:
 
 ```lisp
-(add-hook 'magit-pre-refresh-hook 'diff-hl-magit-pre-refresh)
 (add-hook 'magit-post-refresh-hook 'diff-hl-magit-post-refresh)
 ```
 
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -1000,49 +1000,38 @@ The value of this variable is a mode line template as in
  diff-hl-command-map)
 
 (declare-function magit-toplevel "magit-git")
-(declare-function magit-unstaged-files "magit-git")
+(declare-function magit-git-items "magit-git")
 
-(defvar diff-hl--magit-unstaged-files nil)
-
-(defun diff-hl-magit-pre-refresh ()
-  (unless (and diff-hl-disable-on-remote
-               (file-remote-p default-directory))
-   (setq diff-hl--magit-unstaged-files (magit-unstaged-files t))))
+(define-obsolete-function-alias 'diff-hl-magit-pre-refresh 'ignore "1.11.0")
 
 (defun diff-hl-magit-post-refresh ()
   (unless (and diff-hl-disable-on-remote
                (file-remote-p default-directory))
-   (let* ((topdir (magit-toplevel))
-         (modified-files
-          (mapcar (lambda (file) (expand-file-name file topdir))
-                  (delete-consecutive-dups
-                   (sort
-                    (nconc (magit-unstaged-files t)
-                           diff-hl--magit-unstaged-files)
-                    #'string<))))
-         (unmodified-states '(up-to-date ignored unregistered)))
-    (setq diff-hl--magit-unstaged-files nil)
-    (dolist (buf (buffer-list))
-      (when (and (buffer-local-value 'diff-hl-mode buf)
-                 (not (buffer-modified-p buf))
-                 ;; Solve the "cloned indirect buffer" problem
-                 ;; (diff-hl-mode could be non-nil there, even if
-                 ;; buffer-file-name is nil):
-                 (buffer-file-name buf)
-                 (file-in-directory-p (buffer-file-name buf) topdir)
-                 (file-exists-p (buffer-file-name buf)))
-        (with-current-buffer buf
-          (let* ((file buffer-file-name)
-                 (backend (vc-backend file)))
-            (when backend
-              (cond
-               ((member file modified-files)
-                (when (memq (vc-state file) unmodified-states)
-                  (vc-state-refresh file backend))
-                (diff-hl-update))
-               ((not (memq (vc-state file backend) unmodified-states))
-                (vc-state-refresh file backend)
-                (diff-hl-update)))))))))))
+    (let* ((topdir (magit-toplevel))
+           (modified-files
+            (magit-git-items "diff-tree" "-z" "--name-only" "-r" "HEAD~" "HEAD"))
+           (unmodified-states '(up-to-date ignored unregistered)))
+      (dolist (buf (buffer-list))
+        (when (and (buffer-local-value 'diff-hl-mode buf)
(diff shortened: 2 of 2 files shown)


dgutov:
Hi! Thanks for the PR.

So the basic idea is to switch from `magit-pre-refresh-hook` to `git-commit-setup-hook`?

@tarsius Do you have any thoughts regarding this approach? Any specific pitfalls?

tarsius:
I could see that being an improvement for committing commands, but wouldn't that break the refreshing in other cases, such as "soft reset to HEAD~"?

wyuenho:
I've now changed the names and README to announce that these hooks only apply to commits.

dgutov:
@wyuenho Thank you.

@tarsius Thanks for reviewing. Good point - so that might not work in other scenarios. The post-commit use case is something we can fix first, though, and then see if somebody can improve the general case later.

wyuenho:
@tarsius Do you think `git diff-tree --name-only -r HEAD~ HEAD` is sufficient and fast enough to be called at every `magit-post-refresh-hook` to get a list of changed file names for any scenario? If yes then we can delete the pre-commit hook, rename the post-commit hook back to post-refresh hook.

tarsius:
I think that would be problematic if `HEAD` is a truly large commit.

wyuenho:
Indeed, but the hook is not automatically setup for the users, it always requires an extra step on the user's part to setup, so can we just put a warning in the readme?

dgutov:
It seems to me that creating a large enough commit for this to become a problem should be difficult enough.

And in any case the resources needed to make one will probably overshadow the work we'll need to do in Emacs to process it. I wouldn't bet the house on it, of course, but we could defer solving this problem until somebody reports it.

dgutov:
Speaking of actions that would fall outside of this check, it seems like switching a branch still won't be covered. Maybe we should diff against the previous entry in reflog instead.

ZelphirKaltstahl:
I am just a bystander here and got some e-mails about this PR and read them. A question came up in my mind, that I think might have relevance: The size of a diff ... Did you also consider range diffs? They can more easily get bigger than single commits, since they contain of course usually multiple commits of changes to display.

That's already 2 cents and I do not claim any competency on implementing Emacs packages : )

wyuenho:
> Speaking of actions that would fall outside of this check, it seems like switching a branch still won't be covered. Maybe we should diff against the previous entry in reflog instead.

No need, `magit-post-refresh-hook` has a default value that includes `magit-auto-revert-buffers`, switching branches will run that function. At worst `diff-hl-update` will be called twice on the same buffer, it's probably easier to just fix that [debounce TODO here](https://github.com/dgutov/diff-hl/blob/master/diff-hl.el#L416).

dgutov:
> No need, magit-post-refresh-hook has a default value that includes magit-auto-revert-buffers, switching branches will run that function.

Huh, I suppose that should work. Basically the only case not covered is soft reset.

> At worst diff-hl-update will be called twice on the same buffer, it's probably easier to just fix that

Calling the update 2x is not perfect but probably okay at the first approximation. Debouncing would be nice to have, though.

But from what I see in your last patch, do I get it right that all updates in the "main" scenario also get delayed by 200ms?

wyuenho:
> Basically the only case not covered is soft reset.

This can be covered, the problem is the `vc-state` call reports a cached `up-to-date` after a soft reset. This problem can only be solved if vc could refresh the state of an entire repo instead of on a per-file basis. But if it could do that, we wouldn't need this complicated magit-post-refresh hook to begin with. Perhaps, you could take this request back to emacs-devel and implement a repo-wide refresh in vc in Emacs 31.

> do I get it right that all updates in the "main" scenario also get delayed by 200ms

Only if `diff-hl-update-async` is set to non-nil.

dgutov:
> This problem can only be solved if vc could refresh the state of an entire repo instead of on a per-file basis.

To be clear, I'm not saying we absolutely have to support it. Just enumerating things that might not work - this one seems minor enough.

> Only if diff-hl-update-async is set to non-nil.

This seems suboptimal, however, given that that option was added for a different purpose - and its current users (there probably are more than one) can notice the difference.

wyuenho:
> This seems suboptimal, however, given that that option was added for a different purpose - and its current users (there probably are more than one) can notice the difference.

What purpose? It's always been there for async updates, I just replaced the thread-based update with a debounced update with an user adjustable delay. How else do you debounce? The TODO asked for a debounce, here's a debounce.

wyuenho:
I can try debouncing the thread with a mutex, is this what you are after?

wyuenho:
Turns out there's [magit-inhibit-refresh](https://github.com/magit/magit/blob/e8b85e43d491ca0c0a3fc41ee5d8c4aa34649cd3/lisp/magit-mode.el#L1001), so no need to deal with debouncing

dgutov:
> I can try debouncing the thread with a mutex, is this what you are after?

Debouncing doesn't necessarily imply extra delay - it just means some sort of linearizing the multiple requests which would otherwise overlap.

In any case, not doing that in this PR is okay too. Your last commit still keeps the option `diff-hl-update-async-delay` around. Care to remove it?

wyuenho:
Ok I've delete the last remnant of delayed debounce. I think this implementation is ok for most cases, but reset in general just doesn't work. The fundamental barrier is during reset, the file on disk hasn't really changed unless reset --hard was used, so that doesn't trigger autorevert. In addition, VC is oblivious to the status of the repo if changes to it was not done via VC. What's needed is the ability to poll the git fsmonitor, but then again, I'm not familiar with the IPC messages and they don't seem to be documented well, so I dare not try.

dgutov:
> The fundamental barrier is during reset, the file on disk hasn't really changed unless reset --hard was used, so that doesn't trigger autorevert

Yep, that's it. I wasn't sure if Magit has any tracking of these changes (it feasibly could have, e.g. for of its own invalidation logic on revert, somewhere) - but apparently not. It's okay, still better than before.

> Ok I've delete the last remnant of delayed debounce.

Please also revert the removal of the comment about remote file.

Also we should probably make `diff-hl-magit-pre-refresh` and obsolete-alias to `ignore`, so that the user configs don't break right away after the upgrade.

dgutov:
Thanks! Merging.

PR #233, opened 2025-03-05
Make the maximum width of the fringe indicators customizable

taquangtrung:
Hi,

Currently, the width of the bitmap indicators is the minimum of the fringe width or 16. 

This makes it a bit inflexible to configure the indicator width. For example, I can't configure the diff-hl indicator width to 1 when the fringe width is 8.

So, in this PR, I made the default width 16 customizable, so that the diff-hl indicator width can be flexibly configured.

Can you consider merging it?

Thanks.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -147,6 +147,10 @@
            (set-default var value)
            (when on (global-diff-hl-mode 1)))))
 
+(defcustom diff-hl-default-width 16
+  "Default width of the bitmap indicators."
+  :type 'integer)
+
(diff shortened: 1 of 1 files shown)


dgutov:
Hi!

Maybe the name like `diff-hl-maximum-width` would make more sense?

taquangtrung:
Thanks for the suggestion! 

I updated the code here: https://github.com/dgutov/diff-hl/pull/233/commits/66efb6793c78f241f9ef9091f322980c2ed545b4

dgutov:
Thank you, merging.

PR #236, opened 2025-04-10
Delete temporary file after use

jrosdahl:
I noticed that my `/dev/shm` contains lots of temporary files left by diff-hl. Unless I'm missing something, `rev` in `diff-hl-diff-buffer-with-reference` is always a filename created by `diff-hl-make-temp-file-name`, so delete it after use.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -1141,6 +1141,7 @@ CONTEXT-LINES is the size of the unified diff context, defaults to 0."
            (switches (format "-U %d --strip-trailing-cr" (or context-lines 0))))
       (diff-no-select rev (current-buffer) switches 'noasync
                       (get-buffer-create dest-buffer))
+      (delete-file rev)
       (with-current-buffer dest-buffer
         (let ((inhibit-read-only t))
           ;; Function `diff-sentinel' adds a final line, so remove it


dgutov:
Hi!

By keeping the file around, we get an opportunity to reuse its contents without fetching them again for a while, see the

```
(unless (file-exists-p filename)
```

check in `diff-hl-git-index-revision` and `diff-hl-create-revision`. It might not take too long in the usual case, but still it's nice to save on work.

If the files are annoying, we could put them in a subdirectory. Could also clean them up on a timer, but that's probably premature optimization, given that both /tmp and /dev/shm with be cleared on reboot anyway.

jrosdahl:
I see, thanks for the reply.

Yes, at least to me they are annoying since I quite often look (or tab-complete) in `/dev/shm` only to find thousands of left-over files. Putting them in a subdirectory sounds like a better default if they actually form a cache and are not temporary files. Perhaps also make the location configurable?

dgutov:
Maybe, but maybe it probably won't be used by 99.9% of the users. Just one more var to clutter the customize UI.

What do you think, as someone who actually looks into `/dev/shm/` regularly, is there any potential benefit to using it instead of `/tmp/`, in terms of performance? (StackOverflow disagrees.)

Or, if we did just switch to `/tmp` for that purpose, would having a lot of files directly inside be better?

jrosdahl:
I agree that avoiding unnecessary configure options is good in general. :+1:

Most popular Linux distributions use tmpfs for  `/tmp` as far as I know, so then it's clearly no difference in performance. Not sure about other systems. If `/tmp` is on a disk the short-lived files are cached in the page cache unless the system is low on memory, so I would be surprised if it mattered much even in that case. (Where it potentially differs is if files are `fsync`ed to disk.)

I think that it would be equally unfortunate to "leak" files in `/tmp`.

I now realized that I can add an advice around `diff-hl-make-temp-file-name` to fix my problem. Something like this seems to work:

```elisp
(defvar my-diff-hl-tmpdir "/dev/shm/.diff-hl-cache/")
(unless (file-directory-p my-diff-hl-tmpdir)
  (make-directory my-diff-hl-tmpdir))
(defun my-diff-hl-make-temp-file-name-advice (orig-fun &rest args)
  (let ((temporary-file-directory my-diff-hl-tmpdir))
    (apply orig-fun args)))
(advice-add 'diff-hl-make-temp-file-name :around #'my-diff-hl-make-temp-file-name-advice)
```

It feels like a bit of a hack, though, so it would be nice if the cache was placed in a subdirectory from start. I could look into a PR for that if you like.

I'll close this PR now, though. Thanks for the feedback!

dgutov:
Hmm, does this work?

```diff
diff --git a/diff-hl.el b/diff-hl.el
index fb8ffe2..6337fb4 100644
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -1118,6 +1122,11 @@ the user should be returned."
 
 (declare-function diff-no-select "diff")
 
+(defvar diff-hl-temporary-directory (if (and (eq system-type 'gnu/linux)
+                                             (file-directory-p "/dev/shm/"))
+                                        "/dev/shm/"
+                                      temporary-file-directory))
+
 (defun diff-hl-diff-buffer-with-reference (file &optional dest-buffer backend context-lines)
   "Compute the diff between the current buffer contents and reference in BACKEND.
 The diffs are computed in the buffer DEST-BUFFER. This requires
@@ -1128,10 +1137,7 @@ CONTEXT-LINES is the size of the unified diff context, defaults to 0."
   (save-current-buffer
     (let* ((dest-buffer (or dest-buffer "*diff-hl-diff-buffer-with-reference*"))
            (backend (or backend (vc-backend file)))
-           (temporary-file-directory
-            (if (and (eq system-type 'gnu/linux) (file-directory-p "/dev/shm/"))
-                "/dev/shm/"
-              temporary-file-directory))
+           (temporary-file-directory diff-hl-temporary-directory)
            (rev
             (if (and (eq backend 'Git)
                      (not diff-hl-reference-revision)
```

Your advice seems good, but tmpdir is used in a few more functions, so some files could stay around (not so many, though).

jrosdahl:
That would be fine by me.

PR #237, opened 2025-04-29
Provide diff switches for jj

vifon:
Note that for those specific switches, they are added on top of the original ones as `vc-jj` already sets `vc-jj-diff-switches` to non-nil (the value is `("--git")` by default to be specific).

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -308,6 +308,7 @@ It can be a relative expression as well, such as \"HEAD^\" with Git, or
 
 (defvar vc-svn-diff-switches)
 (defvar vc-fossil-diff-switches)
+(defvar vc-jj-diff-switches)
 
 (defmacro diff-hl-with-diff-switches (body)
   `(let ((vc-git-diff-switches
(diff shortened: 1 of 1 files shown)


rudi:
Hmm, `vc-jj-diff-switches` is a user option, so it's not guaranteed that its value will contain `--git`.  If diff-hl needs specific arguments in order to work, maybe the current value of `vc-jj-diff-switches` should be completely overridden?

vifon:
Yes, you're both right. As per https://codeberg.org/emacs-jj-vc/vc-jj.el/pulls/60, `--git` is now passed internally regardless of `vc-jj-diff-switches`.

dgutov:
Thanks!

vifon:
After some internal discussion and some thinking, I think `--git` should remain in `vc-jj-diff-switches`, and we can include it explicitly in these switches here too. I'll open a new PR in a moment. Apologies for the chaos!

dgutov:
Cool, no problem

PR #239, opened 2025-05-10
Some polish for `diff-hl-show-hunk`

dgutov:
Hi @alvarogonzalezsotillo, please check this out if you have the time and you're still using the features.

I made a pass at the code to make the rendering backends more similar in behavior (wrt positioning), and less "blinky" with the posframe backend.

It removes a couple of workarounds which mention Gnome but I couldn't reproduce (with version 46 anyway) - to fix the side-effects, which both had. Would be great if you could test and see if those were still necessary.

### DIFF
--- a/diff-hl-inline-popup.el
+++ b/diff-hl-inline-popup.el
@@ -154,7 +154,8 @@ FOOTER are showed at start and end."
                  diff-hl-inline-popup--current-footer)))
       ;; https://debbugs.gnu.org/38563, `company--replacement-string'.
       (add-face-text-property 0 (length str) 'default t str)
-      (overlay-put diff-hl-inline-popup--current-popup 'after-string str))))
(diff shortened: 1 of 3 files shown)


dgutov:
Hi @wyuenho could you test out these changes as well? As I see you use the `show-hunk` feature.

This should help me with progressing in the nearby PRs too.

alvarogonzalezsotillo:
Hi @dgutov ,

Im currently away of my computer, so i can not check these changres

I will check them next week, but they look good to me. Feel free to merge them if you need it.

wyuenho:
I've manually applied the patch to #228 with some small adjustments due to naming changes, it works well. However, I don't use the show-hunk-posframe, so I only tested show-hunk-inline. Based on what I see, works well.

dgutov:
Thank you both, merging.

PR #243, opened 2025-07-09
Support a user configurable number of context lines

ideasman42:
Support adjusting the context lines from the default of 3.

----

In some cases I want to stage/revert smaller hunks, this makes it possible.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -214,6 +214,10 @@ the NEW revision is not specified (meaning, the diff is against
 the current version of the file)."
   :type 'boolean)
 
+(defcustom diff-hl-context-lines 3
+  "The number of context lines to include."
+  :type 'integer)
+
 (defcustom diff-hl-update-async nil
   "When non-nil, `diff-hl-update' will run asynchronously.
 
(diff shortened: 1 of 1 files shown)


dgutov:
Perhaps the goal is to alter the contents of the buffer you see after `M-x diff-hl-stage-some`.

In that case, could I suggest trying `C-c C-s` first?

I wouldn't mind adding an option too, but it seems the above command would be more flexible in most cases, and without losing the context which usually helps orient oneself.

ideasman42:
I'm not familier with `diff-hl-split-away-changes` although I'm not sure it makes this pull request redundant, currently I use:

- `diff-hl-revert-hunk`
- `diff-hl-stage-dwin`

As I understand it, in both cases the "current-hunk" will depend on the  number of context lines, so being able to control this allows for more granular control.

dgutov:
> As I understand it, in both cases the "current-hunk" will depend on the number of context lines, so being able to control this allows for more granular control.

The number of context lines could change, but the actual hunk to use is chosen carefully.

Here's an example, there are two hunks separated by just one line of context. I stage just one of them using `diff-hl-stage-current-hunk` (called by `diff-hl-stage-dwim`) with the current settings:

[Screencast from 2025-07-10 05-20-09.webm](https://github.com/user-attachments/assets/cc618feb-80c7-4ea6-adfd-bca1494362ab)

dgutov:
If it works differently for you, please show the steps to reproduce.

PR #246, opened 2025-09-21
Introduce a no-thread async implementation

dgutov:
The main benefit is to avoid current Lisp threading bugs like https://debbugs.gnu.org/79334.

A secondary benefit could be asynchronous calls over Tramp using "direct async", but so far the numbers show no improvement (to test, adjust diff-hl-async-inhibit-functions and do the tramp-direct-async-process stuff).


dgutov:
Reference for TRAMP setup: https://coredumped.dev/2025/06/18/making-tramp-go-brrrr./

PR #247, opened 2025-09-22
Support using reference revision as project-local

EugeneNeuron:
Hi, this PR makes `diff-hl-reference-revision` buffer-local and project-local to avoid a global setting.

To make it project-local I use `dir-local` in-memory functionality:
- Newly opened files of the same project receive up-to-date `diff-hl-reference-revision` value from `dir-local`.
- To update this value in already open files we traverse buffers of the current project only (previously all buffers).

All tests pass (`emacs -Q`).

* diff-hl.el (diff-hl-reference-revision): Make it local and safe if string or nil.
(diff-hl-set-reference-rev): Make reference revision project-local. 
(diff-hl-reset-reference-rev): When called, reset reference revision for the current project only, with when called with the prefix argument, reset it globally.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -270,11 +270,21 @@ and passed the value `default-directory'.
 If any returns non-nil, `diff-hl-update' will run synchronously anyway."
   :type '(repeat :tag "Predicate" function))
 
+(defvar diff-hl-reference-revision-projects-cache '()
+  "Alist of cached directory roots for per-project reference revisions.
+Each element in this list has the form (DIR . REV).
+DIR is the expanded name of the directory.
+REV is the current reference revision.")
+
 (defvar diff-hl-reference-revision nil
   "Revision to diff against.  nil means the most recent one.
 
 It can be a relative expression as well, such as \"HEAD^\" with Git, or
 \"-2\" with Mercurial.")
+
+(put 'diff-hl-reference-revision 'safe-local-variable
+     (lambda (value)
+       (or (null value) (stringp value))))
 
 (defun diff-hl-define-bitmaps ()
   (let* ((scale (if (and (boundp 'text-scale-mode-amount)
@@ -1257,7 +1267,13 @@ The value of this variable is a mode line template as in
         ;; Magit versions 2.0-2.3 don't do the above and call this
         ;; instead, but only when they don't call `revert-buffer':
         (add-hook 'magit-not-reverted-hook 'diff-hl-update nil t)
-        (add-hook 'text-scale-mode-hook 'diff-hl-maybe-redefine-bitmaps nil t))
+        (add-hook 'text-scale-mode-hook 'diff-hl-maybe-redefine-bitmaps nil t)
+        (when-let* ((rev (map-some
+                          (lambda (root rev)
+                            (when (string-prefix-p root default-directory)
+                              rev))
+                          diff-hl-reference-revision-projects-cache)))
+          (setq-local diff-hl-reference-revision rev)))
     (remove-hook 'after-save-hook 'diff-hl-update t)
     (remove-hook 'after-change-functions 'diff-hl-edit t)
     (remove-hook 'find-file-hook 'diff-hl-update-once t)
@@ -1266,7 +1282,8 @@ The value of this variable is a mode line template as in
     (remove-hook 'magit-not-reverted-hook 'diff-hl-update t)
     (remove-hook 'text-scale-mode-hook 'diff-hl-maybe-redefine-bitmaps t)
     (diff-hl-remove-overlays)
-    (diff-hl--autohide-margin)))
+    (diff-hl--autohide-margin)
+    (kill-local-variable 'diff-hl-reference-revision)))
 
 (defun diff-hl-after-checkin ()
   (let ((fileset (vc-deduce-fileset t)))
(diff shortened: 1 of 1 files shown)


EugeneNeuron:
OK, I can do it.

Global `diff-hl-reset-reference-rev` doesn't make any sense between any two different repos, AFAIK.
The only situation where global `diff-hl-reset-reference-rev` might be appropriate is worktrees of the same repo (`git worktree`).

Just out of curiosity, is that the only reason you'd prefer to keep the old behavior?

EugeneNeuron:
@dgutov  I need a few inputs from, please:
1. If `diff-hl-set-reference-rev` is called from a buffer with project-local value, should it change: 
   1. global
   2. global + local value of the current project (project-local is removed or updated?)
   3. local value of the current project
2. May I keep the ability to reset the reference revision when `diff-hl-set-reference-rev` is called with prefix?
   - If yes, then what exactly should that reset: 
     1. global, 
     2. local of the project at hand
     3. global+all local?
3. What exactly should `diff-hl-reset-reference-rev` reset: 
   1. global, 
   2. local of the project at hand 
   3. global+all local?

My personal opinion:
1. global only
Since in this situation the user called `-in-project` command before, they know about the difference and shouldn't confuse them.
2. Yes, because it won't hurt backward compatibility - previously there was no reason to call this command with the prefix. 
If it's called with one prefix - reset global only, if called with two prefixes - reset global and remove all project-local values (make it global in those buffers).
3. global only when called without the prefix and global + project-local when called with the prefix (makes it global in those buffers as well)

diff-hl.el
```diff
@@ -1525,42 +1534,119 @@ view buffer.
...
+     (current-prefix-arg
+      (diff-hl-set-reference-rev-in-project-internal nil)
+      (message "Reference revision reset to the most recent revision (project %s)" name))
+     ;; set
+     (t (diff-hl-set-reference-rev-in-project-internal rev)
+        (message "Showing changes against %s (project %s)" rev name)))))
+
+(defun diff-hl-set-reference-rev-in-project-internal (rev)
+  (let* ((proj (project-current))
+         (class (intern (concat "diff-hl-" (project-name proj)))))
```
catern, line 1618: project-name can easily conflict, you probably should use (some variation on) project-root instead?

diff-hl.el
```diff
@@ -1525,42 +1534,119 @@ view buffer.
...
+    ;; newly opened files will share this value
+    (dir-locals-set-class-variables class
+                                    `((nil (diff-hl-reference-revision . ,rev))))
+    ;; run once per project
+    (unless (assq class dir-locals-directory-cache)
+      (dir-locals-set-directory-class (project-root proj)
+                                      class))
```
catern, line 1625: @dgutov shouldn't this be something project.el handles, if we want to have project-local variables implemented this way?

catern (approved):

diff-hl.el
EugeneNeuron: Thanks, indeed.
Switched to `project-root` directly and made sure that symbols with ~,/ and spaces can be interned.
Symbol doesn't look fancy (e.g. `diff-hl-~/git/diff-hl/`), but it's never exposed to user and works.

diff-hl.el
dgutov: Good question.

@EugeneNeuron Do you think that instead we would want to use `add-dir-local-variable`, to alter (or create) a corresponding `.dir-locals.el` file, adding such a variable there?

This can have a benefit that a) it persists between sessions, b) you can fairly easily find out which value is used for which project, instead of trying introspection on the values altered by `dir-locals-set-directory-class`.

diff-hl.el
```diff
@@ -1525,42 +1534,119 @@ view buffer.
...
-  (setq diff-hl-reference-revision rev)
-  (dolist (buf (buffer-list))
-    (with-current-buffer buf
-      (when diff-hl-mode
-        (when (bound-and-true-p diff-hl-amend-mode)
-          (diff-hl-amend-mode -1))
-        (diff-hl-update)))))
+  (unless current-prefix-arg
+    (message "Set global reference revision to %s" rev))
+  (cond ((equal '(4) current-prefix-arg)
```
dgutov, line 1559: Please add a newline after `(cond`.

dgutov:
> It's a pity that there's no built-in method to make a variable project-wide yet, but I'll explain why I need it project-wide

This totally makes sense. It's just the most obvious way to create a directory-local value is to add it to a dir-locals file. Most obvious to certain long-times Emacs users.

But I'm fine with however you want to implement it, FWIW.

diff-hl.el
```diff
@@ -1257,7 +1267,12 @@ The value of this variable is a mode line template as in
         ;; Magit versions 2.0-2.3 don't do the above and call this
         ;; instead, but only when they don't call `revert-buffer':
         (add-hook 'magit-not-reverted-hook 'diff-hl-update nil t)
-        (add-hook 'text-scale-mode-hook 'diff-hl-maybe-redefine-bitmaps nil t))
+        (add-hook 'text-scale-mode-hook 'diff-hl-maybe-redefine-bitmaps nil t)
+        (when-let* ((proj (project-current))
```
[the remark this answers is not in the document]
EugeneNeuron: Adjusted logic to both remarks.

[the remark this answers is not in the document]
dgutov: Thanks! Merging.

I'll fixup the build on 27/28 after.

PR #248, opened 2025-09-24
Fix `diff-hl-resolved-reference-revision` for Hg and Bzr with buffer-local `diff-hl-reference-revision`

whhone:
**Problem**:
`diff-hl-resolved-reference-revision` fails to work correctly when `diff-hl-reference-revision` is set as a buffer-local variable.

Example:
```emacs-lisp
(setq-local diff-hl-reference-revision "test-rev")
```

**Cause**:
When the implementation for backends like Hg (and similarly Bzr) uses `with-temp-buffer`, buffer-local variables from the original buffer context are not accessible. Consequently, `diff-hl-reference-revision` evaluates to `nil` within the temporary buffer.

For 'Hg as an exmaple,
```emacs-lisp
(with-temp-buffer
  (vc-hg-command (current-buffer) 0 nil
                 "identify" "-r" diff-hl-reference-revision
                 "-i")
  ...)
```

This leads to an incorrect command being executed:
*   **Expected hg command**: `hg identify -r test-rev -i`
*   **Actual hg command**: `hg identify -r -i`

**Fix**:
To resolve this, the value of `diff-hl-reference-revision` is now copied into a non-buffer-local variable (e.g., `original-diff-hl-reference-revision`) *before* entering the `with-temp-buffer` block. This non-buffer-local variable is then used within the temporary buffer context, ensuring the correct revision is always passed to the underlying version control command.

**Tests**:
I have tested it in my hg repo. Also added a test in `diff-hl-test.el`.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -1442,21 +1442,35 @@ CONTEXT-LINES is the size of the unified diff context, defaults to 0."
    ((eq backend 'Git)
     (vc-git--rev-parse diff-hl-reference-revision))
    ((eq backend 'Hg)
-    (with-temp-buffer
-      (vc-hg-command (current-buffer) 0 nil
-                     "identify" "-r" diff-hl-reference-revision
-                     "-i")
-      (goto-char (point-min))
-      (buffer-substring-no-properties (point) (line-end-position))))
+    ;;  Preserve the potentially buffer-local variables (e.g.,
+    ;;  `diff-hl-reference-revision`, `vc-hg-program`) by not switching buffer
+    ;;  until the vc-hg-command is done.
+    (let ((temp-buffer (generate-new-buffer " *temp*")))
+      (unwind-protect
+          (progn
+            (vc-hg-command temp-buffer 0 nil
+                           "identify" "-r" diff-hl-reference-revision "-i")
+            (with-current-buffer temp-buffer
+              (goto-char (point-min))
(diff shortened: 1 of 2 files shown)


diff-hl.el
```diff
@@ -1442,19 +1442,25 @@ CONTEXT-LINES is the size of the unified diff context, defaults to 0."
...
    ((eq backend 'Hg)
-    (with-temp-buffer
-      (vc-hg-command (current-buffer) 0 nil
-                     "identify" "-r" diff-hl-reference-revision
-                     "-i")
-      (goto-char (point-min))
-      (buffer-substring-no-properties (point) (line-end-position))))
+    ;;  Preserve the potentially buffer-local `diff-hl-reference-revision`
+    ;;  inside the `with-temp-buffer`.
+    (let ((original-diff-hl-reference-revision diff-hl-reference-revision))
```
dgutov, line 1447: Let's just call it `reference-revision`?

whhone: Done.

whhone:
I update the PR to cover all buffer-local variables.

dgutov:
Ah now I get it - thank you. The major difference is that the buffer switch only happens later.

The test is not how I'd really write it, but it shouldn't hurt either, so I'm merging like this now.

whhone:
Thanks for reviewing! @dgutov 

The test is also not how I'd write it, but it's the best I can do with the time I have. :)

PR #253, opened 2025-10-17
Improve interaction with Jj and change fn to diff-hl-resolved-revision

### LINKED ISSUE #250: diff-hl-show-hunk* doesn't work correctly with jj
Most `diff-hl` commands work with `vc-jj` out of the box, e.g. `diff-hl-diff-goto-hunk` or `diff-hl-next-hunk`, however, the family of `diff-hl-show-hunk*` functions do not. The message says `"No next hunk"` or `"There is no next change"`. This is probably because `diff-hl` compares with the `@` revision (current revision) rather than `@-`. If I modify `diff-hl-reference-revision` from `nil` to `"@-"`, showing hunk with JJ seems to work correctly.

dgutov:
Fixes #250

### DIFF
--- a/test/diff-hl-test.el
+++ b/test/diff-hl-test.el
@@ -221,20 +221,20 @@
 Diff finished."
                  (buffer-substring (point) (point-max))))))))
 
-(diff-hl-deftest diff-hl-resolved-reference-revision-buffer-local-hg ()
+(diff-hl-deftest diff-hl-resolved-revision-with-hg ()
   (diff-hl-test-in-source
-   (setq-local diff-hl-reference-revision "test-rev")
-   (setq-local vc-hg-program "chg")
-   (condition-case err
-       (progn
-         (diff-hl-resolved-reference-revision 'Hg)
-         (ert-fail "Expected an error to be signaled but none was."))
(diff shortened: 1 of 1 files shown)


dgutov:
@whhone Hi!

Aside from test failures, does this change look good for you? Previously you submitted this code change in https://github.com/dgutov/diff-hl/pull/248.

I think the proposed keeps buffer-local `diff-hl-reference-revision` working, but if you had any other such vars which would stop working, please chime in.

dgutov:
I can revert that part, if that's needed.

Something of note, however, is that the Git implementation, and the newly added Jj one, don't support buffer-local values for their corresponding programs. Should we fix that too?

Maybe we could come up with a shared solution.

whhone:
I believe JJ and git already respect buffer-local variables since they do not switch to a temp buffer.

For example, you can try setting `vc-git-program` or `vc-jj-program` as a buffer-local variable and see if they are being respected.

whhone:
Ops... I did not go into the implementation detail of `vc-git--rev-parse` and `vc-jj--process-lines`.

I don't have strong opinion on respecting respecting `buffer-local` variables or not. I can find workaround in my case. Thanks!

PR #254, opened 2025-10-31
fix(diff-hl.el): Silence compiler warnings

bcardoso:
Fix compiler warnings:

diff-hl.el:1677:19: Warning: the function ‘project-buffers’ is not known to be defined.
diff-hl.el:1668:9: Warning: the function ‘project-roots’ is not known to be defined.
diff-hl.el:1654:17: Warning: the function ‘project-name’ is not known to be defined.


dgutov:
Thanks!

PR #255, opened 2025-11-02
Fix compiler warnings

tarsius:
I've implemented these fixes several weeks ago but did not submit them because of the work involved justifying the changes.  I am now submitting them anyway because another attempt to fix[^1] some of these warnings was just merged, and that approach was wrong.

See the individual commit for more information.

[^1]: #254 /cc @bcardoso


bcardoso:
Thanks for the detailed explanation, Jonas. Sorry for my naive approach.

PR #257, opened 2025-11-02
fix #256: prevent diff-hl-inline-popup-hide from being called twice

### LINKED ISSUE #256: inline-popup crashes when clicking on a second hunk before quitting the current one
This is tested with master.

I have `diff-hl-flydiff-mode` and `diff-hl-margin-mode` turned on, and `debug-on-error` turned on as well.

https://github.com/user-attachments/assets/9912c7a1-90d6-402b-9abd-738fb401b915

```
Debugger entered--Lisp error: (wrong-type-argument overlayp nil)
  overlay-get(nil diff-hl-hunk-type)
  (let* ((lines (split-string (save-current-buffer (set-buffer buffer) (buffer-string)) "[\n\15]+")) (smart-lines diff-hl-show-hunk-inline-popup-smart-lines) (original-lines-number (cl-count-if #'(lambda (s) (string-prefix-p "-" s)) lines)) (lines (if (string= (car (last lines)) "") (butlast lines) lines)) (lines (if (and (eq original-lines-number 0) smart-lines) diff-hl-show-hunk--no-lines-removed-message lines)) (overlay diff-hl-show-hunk--original-overlay) (type (overlay-get overlay 'diff-hl-hunk-type)) (point (if (eq type 'delete) (overlay-start overlay) (overlay-end overlay))) (propertize-line #'(lambda (l) (propertize l 'face (cond (... ...) (... ...))))) (propertized-lines (mapcar propertize-line lines))) (save-excursion (if (and diff-hl-show-hunk-inline-popup-hide-hunk (not (eq type 'delete))) (progn (let* ((invisible-overlay (make-overlay ... ...))) (overlay-put invisible-overlay 'invisible t) (setq diff-hl-show-hunk--hide-function #'(lambda nil ... ... ...))))) (diff-hl-show-hunk--goto-hunk-overlay overlay) (let ((height (if smart-lines (progn (if ... ...)))) (footer "(q)Quit  (p)Previous  (n)Next  (r)Revert  (c)Copy original")) (if diff-hl-show-staged-changes nil (setq footer (concat footer " (S)Stage"))) (diff-hl-inline-popup-show propertized-lines (if (and (boundp 'diff-hl-reference-revision) diff-hl-reference-revision) (concat "Diff with " diff-hl-reference-revision) "Diff with HEAD") footer diff-hl-show-hunk-map #'diff-hl-show-hunk-hide point height))))
  diff-hl-show-hunk-inline-popup(#<buffer *diff-hl-show-hunk-buffer*> 7)
  diff-hl-show-hunk()
  diff-hl-show-hunk--click((mouse-1 (#<window 18 on init.el> left-margin (3 . 590) 256185107 (#(" " 0 1 (face diff-hl-margin-change)) . 0) 65392 (0 . 36) nil (3 . 12) (7 . 16))))
  funcall-interactively(diff-hl-show-hunk--click (mouse-1 (#<window 18 on init.el> left-margin (3 . 590) 256185107 (#(" " 0 1 (face diff-hl-margin-change)) . 0) 65392 (0 . 36) nil (3 . 12) (7 . 16))))
  command-execute(diff-hl-show-hunk--click)
```

The problem seems to be related to `(diff-hl-inline-popup-hide)` being called twice inside `(diff-hl-show-hunk-inline-popup)` because `diff-hl-inline-popup--close-hook` is also set to `diff-hl-show-hunk-hide` on the first run, which is just going to delegate back to `diff-hl-inline-popup-hide`.

wyuenho:
Fixes https://github.com/dgutov/diff-hl/issues/256

### DIFF
--- a/diff-hl-show-hunk.el
+++ b/diff-hl-show-hunk.el
@@ -232,7 +232,9 @@ Returns a list with the buffer and the line number of the clicked line."
 (defun diff-hl-show-hunk-inline-popup (buffer &optional _ignored-line)
   "Implementation to show the hunk in a inline popup.
 BUFFER is a buffer with the hunk."
-  (diff-hl-inline-popup-hide)
+  ;; prevent diff-hl-inline-popup-hide from being called twice
+  (let ((diff-hl-inline-popup--close-hook nil))
+    (diff-hl-inline-popup-hide))
   (setq diff-hl-show-hunk--hide-function #'diff-hl-inline-popup-hide)
   (let* ((lines (split-string (with-current-buffer buffer (buffer-string)) "[\n\r]+" ))
          (smart-lines diff-hl-show-hunk-inline-popup-smart-lines)


dgutov:
Seems like a good idea. Thanks!

BTW I think `diff-hl-show-hunk-mouse-mode` should have been mentioned among the prerequisites.

PR #260, opened 2025-11-20
Add option to auto recenter after diff-hl-{next,previous}-hunk

NicholasBHubbard:
Hello. This PR adds a new variable `diff-hl-next-previous-hunk-auto-recenter`, that when set causes `diff-hl-next-hunk` and `diff-hl-previous-hunk` to automatically call `recenter` after moving the point.

In my configuration I use after advice on these functions to do this auto-recentering, but I thought it would be nicer if there was a diff-hl variable for this behavior. 

Let me know if you think this is a good idea, and if there is anything you would like changed in this PR. Thanks!

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -142,6 +142,11 @@ enclosed in a `progn' form.  ELSE-FORMS may be empty."
   :group 'diff-hl
   :type 'boolean)
 
+(defcustom diff-hl-next-previous-hunk-auto-recenter nil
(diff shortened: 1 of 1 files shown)


NicholasBHubbard:
Any update on this? Do others think this would be a good addition to `diff-hl`?

frou:
> Do others think this would be a good addition to `diff-hl`?

Yeah - I think it's a good idea

dgutov:
Thanks for the reminder and the feedback! Merging :rocket:

PR #262, opened 2025-12-07
diff-hl--resolve: Use vc-run-delayed-success on Emacs >=31

spwhitton:
Eventually this can replace diff-hl's custom replacement for vc-exec-after.


oscarfv:
@spwhitton : thoughts on #263 ?

PR #266, opened 2026-01-04
Add Ediff support

ultronozm:
Add the command 'diff-hl-ediff-current-hunk', bound to 'C-x v e' in
'diff-hl-mode-map' and 'e' in the show-hunk UI, that uses Ediff to compare
against the reference revision, jumping to the hunk nearest point.

* diff-hl.el (diff-hl--use-git-index-base-p): New helper function.
(diff-hl-diff-against-reference, diff-hl-diff-buffer-with-reference): Use it.
(diff-hl--ediff-reference-buffer): New helper function.
(diff-hl-ediff-current-hunk): New command.
(diff-hl-command-map): Bind 'e' to 'diff-hl-ediff-current-hunk'.
(diff-hl-repeat-exceptions): Add 'diff-hl-ediff-current-hunk'.

* diff-hl-show-hunk.el (diff-hl-show-hunk-ediff): New command.
(diff-hl-show-hunk-map): Bind it to 'e'.

* diff-hl-show-hunk-inline.el (diff-hl-show-hunk-inline): Update footer text to
include Ediff keybinding.

* diff-hl-show-hunk-posframe.el (diff-hl-show-hunk-posframe--header-line): Add
Ediff button to the posframe header line.

### DIFF
--- a/diff-hl-show-hunk-inline.el
+++ b/diff-hl-show-hunk-inline.el
@@ -361,7 +361,7 @@ BUFFER is a buffer with the hunk."
              (when smart-lines
                (when (not (eq 0 original-lines-number))
                  original-lines-number)))
-            (footer "(q)Quit  (p)Previous  (n)Next  (r)Revert  (c)Copy original"))
+            (footer "(q)Quit  (p)Previous  (n)Next  (e)Ediff  (r)Revert  (c)Copy original"))
         (unless diff-hl-show-staged-changes
           (setq footer (concat footer " (S)Stage")))
         (diff-hl-show-hunk-inline-show
--- a/diff-hl-show-hunk-posframe.el
+++ b/diff-hl-show-hunk-posframe.el
@@ -146,6 +146,11 @@ The button calls an ACTION."
     #'diff-hl-show-hunk-copy-original-text)
 
    (diff-hl-show-hunk--posframe-button
+    "⇄ Ediff"
+    "Ediff (\\[diff-hl-show-hunk-ediff])"
+    #'diff-hl-show-hunk-ediff)
+
+   (diff-hl-show-hunk--posframe-button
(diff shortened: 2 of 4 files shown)


ultronozm:
I just rebased this PR onto the latest master. Anything I could do to make it easier to review? For what it's worth, it's a feature that I've been using practically every day.

dgutov:
@ultronozm Hi! Thanks for the ping.

It's a notable addition. I'm not an ediff user myself, so it will take a bit of time, for me to understand how it fits the existing workflows.

ultronozm:
In case it helps, here's the user-facing story as I see it.

`diff-hl-mode` shows which parts of a buffer have changed. `C-x v [` and `C-x v ]` make it easy to jump between changed regions.

For a changed region, `C-x v *` shows the corresponding diff hunk, and `C-x v n` reverts that hunk.

The proposed command, `C-x v e`, is the heavier-weight version of that same workflow: it opens Ediff on the old and current versions, positioned at the current hunk.  This covers the same inspect/revert use case as `C-x v *` and `C-x v n`, but with Ediff's comparison and copying commands available when the hunk needs closer review.

For instance, suppose you want to partially revert the current hunk by bringing back some, but not all, of the old text.  The Ediff workflow makes this easy: compare the old and new buffers side by side, copy selected differences from the old buffer to the new buffer, or edit the new buffer while keeping the old buffer visible.

dgutov:
Hi! Looked it over again - the design is fine.

I think we'll just need to limit the addition of key binding to the main keymap - it order hot to hijack any more importnt keys from VC.

ultronozm:
Thanks, glad to hear!  I've pushed an update that removes the key binding from the main keymap.

dgutov:
Thanks!

PR #276, opened 2026-04-29
Add support for indirect buffers

### LINKED ISSUE #272: Does this work with indirect buffers?
I use narrow indirect buffers heavily and decided to try diff-hl. It didn't work for my narrow-indirect buffers, but works perfectly for the source file. Investigating further, I tried separately narrowing the buffer and making an indirect buffer. The former worked, while the latter failed.

Does this package support indirect buffers?

jamescherti:
Here is a first attempt to fix the indirect buffer issue: https://github.com/dgutov/diff-hl/issues/272#issuecomment-4320407818

I replaced all instances of `buffer-file-name` with a function that supports indirect buffers:
- Added the `diff-hl--buffer-file-name` function to retrieve the file name of the current buffer or its base buffer.
- Replaced direct accesses of the `buffer-file-name` variable with the new helper changes to all live buffers sharing the same base buffer.

It successfully returns the correct file name. However, it does not fix the issue of the margin and fringe not updating. An additional change is required to fix https://github.com/dgutov/diff-hl/issues/272#issuecomment-4320407818 .

Let us collaborate to address the issue, given your familiarity with the `diff-hl` codebase and your ability to assist or contribute to fixing it once and for all.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -876,7 +876,7 @@ Return a list of line overlays used."
 With double prefix argument (C-u C-u), the diff is made against the
 reference revision."
   (interactive (list current-prefix-arg))
-  (with-current-buffer (or (buffer-base-buffer) (current-buffer))
+  (with-current-buffer (current-buffer)
     (if (equal historic '(16))
         (diff-hl-diff-reference-goto-hunk)
       (diff-hl-diff-goto-hunk-1 historic nil))))
@@ -884,7 +884,7 @@ reference revision."
 (defun diff-hl-diff-reference-goto-hunk ()
   "Run VC diff command against the reference and go to the corresponding line."
   (interactive)
-  (with-current-buffer (or (buffer-base-buffer) (current-buffer))
+  (with-current-buffer (current-buffer)
     (diff-hl-diff-goto-hunk-1 nil diff-hl-reference-revision)))
 
 (defun diff-hl-root-diff-reference-goto-hunk ()
@@ -893,7 +893,7 @@ And if the current buffer is visiting a file, and it has changes, the diff
 buffer will show the position corresponding to its current line."
   (interactive)
   (defvar vc-sentinel-movepoint)
-  (with-current-buffer (or (buffer-base-buffer) (current-buffer))
+  (with-current-buffer (current-buffer)
     (let ((backend (vc-deduce-backend))
           (default-directory default-directory)
           rootdir fileset
@@ -1099,7 +1099,7 @@ its end position."
 (defun diff-hl-revert-hunk ()
   "Revert the diff hunk with changes at or above the point."
   (interactive)
-  (with-current-buffer (or (buffer-base-buffer) (current-buffer))
+  (with-current-buffer (current-buffer)
     (diff-hl-revert-hunk-1)))
 
 (defun diff-hl-hunk-overlay-at (pos)
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -314,6 +314,15 @@ It can be a relative expression as well, such as \"HEAD^\" with Git, or
      (lambda (value)
        (or (null value) (stringp value))))
 
+(defun diff-hl--buffer-file-name (&optional buffer)
+  "Return the file name of the BUFFER or its base buffer.
+BUFFER defaults to the current buffer."
+  (let* ((buffer (or buffer (current-buffer)))
+         (base-buffer (buffer-base-buffer buffer)))
+    (if base-buffer 
+        (buffer-file-name base-buffer)
+      (buffer-file-name buffer))))
+
 (defun diff-hl-define-bitmaps ()
   (let* ((scale (if (and (boundp 'text-scale-mode-amount)
                          (numberp text-scale-mode-amount))
@@ -485,7 +494,7 @@ It can be a relative expression as well, such as \"HEAD^\" with Git, or
   buffer)
 
 (defun diff-hl-changes ()
-  (let* ((file buffer-file-name)
+  (let* ((file (diff-hl--buffer-file-name))
          (backend (vc-backend file))
          (hide-staged (and (eq backend 'Git) (not diff-hl-show-staged-changes))))
     (when backend
@@ -530,7 +539,7 @@ It can be a relative expression as well, such as \"HEAD^\" with Git, or
   (or (assoc-default backend diff-hl-head-revision-alist)
       ;; It's usually cached already (e.g. for mode-line).
       ;; So this is basically an optimization for rare cases.
-      (vc-working-revision buffer-file-name backend)))
+      (vc-working-revision (diff-hl--buffer-file-name) backend)))
 
 (defun diff-hl-adjust-changes (old new)
   "Adjust changesets in OLD using changes in NEW.
@@ -893,8 +902,9 @@ buffer will show the position corresponding to its current line."
           (setq rootdir (vc-call-backend backend 'root default-directory)
                 default-directory rootdir
                 fileset `(,backend (,rootdir))
-                relname (if buffer-file-name (file-relative-name buffer-file-name
-                                                                 rootdir)))
+                relname (let ((file (diff-hl--buffer-file-name)))
+                          (when file
+                            (file-relative-name file rootdir))))
         (error "Directory is not version controlled"))
       (setq fileset (or fileset (vc-deduce-fileset)))
       (vc-buffer-sync-fileset fileset t)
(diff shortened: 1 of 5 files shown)


dgutov:
Try removing the `(with-current-buffer (or (buffer-base-buffer) (current-buffer))` wrappers too, there are 4 of them.

jamescherti:
I replaced `(with-current-buffer (or (buffer-base-buffer) (current-buffer))` with `(current-buffer)` in the latest commit, but it doesn't seem to fix the issue.

jamescherti:
I found a way to fix it. Try the latest commit.

I updated `diff-hl--update` to synchronize overlays across indirect buffers. The latest commit makes `diff-hl--update` identify the base buffer of the current file and iterate through all live buffers to apply diff highlighting to any indirect clones.

jamescherti:
We could optimize this by making `diff-hl` track indirect buffers to avoid using `buffer-list`. Please read and test the latest commit and send me your feedback.

dgutov:
> but it doesn't seem to fix the issue

When you said "the issue", did you mean that the current indirect buffer was not updated after editing and saving? Or that some other related buffers were not updated.

jamescherti:
The issue: `diff-hl-mode` fails to display or update version control indicators within indirect buffers when the base buffer is modified, or within base buffers when the indirect buffer is modified.

Applying this patch ensures that `diff-hl` highlights correctly synchronize and render across both the base buffer and any associated indirect buffers whenever file changes are saved.

dgutov:
I'm asking specifically about the last commit. It seems to be more in the "nice to have" category. Iterating along the whole buffer list (or managing the indirect buffers list) has a cost as well. So how about we clarify that first.

jamescherti:
I can optimize this by storing a list of indirect buffers in a local variable in the base buffer each time an indirect buffer with `diff-hl-mode` enabled is created or when `diff-hl-mode` is started. This way, instead of looping over all buffers, the function will only loop over the indirect buffers and the base buffer.

Please confirm that:

1. You have tested it and it works for you too.
2. You are interested in merging this pull request if I implement it.

This way, I will not spend time on it unnecessarily.

dgutov:
Could we get the answer to my last question first?

If the last change is less critical, we could split the improvement this way:

* Simpler change that only updates the current buffer (whether it's indirect or base).
* New hook that allows you to set up the update of related buffers.

Or something like that, I haven't decided yet. You don't need to polish the implementation further yet, but testing the alternatives and describing the results (and possible downsides) is what's valuable here.

jamescherti:
Hi Dmitry,

I am currently using this implementation with many indirect buffers and it works well. This pull request can be merged as is after more testing. (The `buffer-list` loop is fast because it skips buffers using `(eq (buffer-base-buffer buf) base)`, and as I mentioned, it can easily be optimized in the future.)

To answer your questions: the last change is not just an optional enhancement. It is required because without it, the highlighting does not work at all in indirect buffers.

While my initial commits fixed the file name resolution, the margin and fringe still failed to update or display any indicators in the indirect buffer. The logic introduced in the last commit is what actually ensures the visual elements to render in the indirect buffers/clones.

Splitting the improvement into a simpler change for the active buffer and a separate hook for related buffers will not resolve the issue.

Just try it with and without the latest commit. You'll see the difference.

jamescherti:
Hello @dgutov, 

Just checking in on this PR. Let me know if any changes are needed or if there is anything blocking review.

I've been using this for many days and it is working well.

Thanks.

diff-hl.el
```diff
@@ -1424,26 +1441,26 @@ The value of this variable is a mode line template as in
             (magit-git-items "diff-tree" "-z" "--name-only" "-r" "HEAD~" "HEAD"))
            (unmodified-states '(up-to-date ignored unregistered)))
       (dolist (buf (buffer-list))
-        (when (and (buffer-local-value 'diff-hl-mode buf)
-                   (not (buffer-modified-p buf))
-                   ;; Solve the "cloned indirect buffer" problem
-                   ;; (diff-hl-mode could be non-nil there, even if
-                   ;; buffer-file-name is nil):
-                   (buffer-file-name buf)
-                   (file-in-directory-p (buffer-file-name buf) topdir)
```
dgutov, line 1435: As a matter of style, I think you can avoid reindenting and just call the helper twice - IIUC it's fast enough that it wouldn't matter for performance.

(Just in this function.)

dgutov:
@jamescherti Hi, sorry about the wait.

What was missing for me is the explanation "why". So apparently when an indirect buffer is saved, that happens in the base buffer, and the after-save-hook is called only in there too, not in the current (indirect) buffer. Unless the flydiff mode is used, then this issue does not apply.

Anyway, that implies that all the changes are necessary indeed. Too many places for my liking (for a minor feature) but oh well.

dgutov:
See inline comments (minor)

jamescherti:
I resolved your comments.

I had to resolve conflicts because it has been a long time. I think it is better to merge this so we can avoid conflicts in the future.

jamescherti:
> What was missing for me is the explanation "why". So apparently when an indirect buffer is saved, that happens in the base buffer, and the after-save-hook is called only in there too, not in the current (indirect) buffer. Unless the flydiff mode is used, then this issue does not apply.

Here is why:

1. When an indirect buffer is saved, `diff-hl` highlights are not updated in the base buffer.
2. When a base buffer is saved, `diff-hl` highlights are not updated in the indirect buffers.

> Anyway, that implies that all the changes are necessary indeed. Too many places for my liking (for a minor feature) but oh well.

This is not a minor feature. For many users, indirect buffers is as much a part of their Emacs workflow as narrowing.

dgutov:
Thanks.

The second comment is still pending.

jamescherti:
My pleasure.

However, I only see one comment, which I marked as resolved.

image

dgutov:
Probably just github being buggy lately: 
image

jamescherti:
I've added the `diff-hl--target-buffer` helper function that returns the correct buffer for the situation.

dgutov:
Thanks! Time to merge.

@jamescherti Speaking of iterating across the whole buffer list, though, what would you say if only the currently visible indirect buffers were updated? Or something to that effect.

So that the default behavior is fast no matter how many buffers a user has (sometimes people do visit 1000+ buffers, I remember bug reports with scenarios like that).

jamescherti:
Hello @dgutov,

Thank you for merging this pull request! This is a step in the right direction, allowing enhancing the indirect buffer implementation with minimal merge conflicts. I use indirect buffers heavily in my workflow and will test these changes thoroughly. If I encounter any issues, I will submit a new pull request.

The current implementation will not be slow due to the number of buffers. The specific part of the function that interacts with your open buffers is this loop:
```elisp
(dolist (buf (buffer-list))
  (when (and (buffer-live-p buf)
             (eq (diff-hl--target-buffer buf) base)
             (buffer-local-value 'diff-hl-mode buf))
```

While (buffer-list) will return a list of 1000 items, iterating through them is an O(N) operation that does very little work per iteration. For each buffer, it evaluates three fast predicates: buffer-live-p, pointer equality with eq, and buffer-local-value. Executing these checks 1000 times takes a fraction of a millisecond in Emacs Lisp. The heavy operations, such as diff-hl-remove-overlays and diff-hl–update-overlays, are gated behind `(eq (diff-hl–target-buffer buf) base)`. This check ensures the overlay logic only runs for the current base buffer and its indirect clones. Unless you have hundreds of indirect buffers open for the exact same file simultaneously, that inner block will only execute once or twice.

dgutov:
> While (buffer-list) will return a list of 1000 items, iterating through them is an O(N) operation that does very little work per iteration. For each buffer, it evaluates three fast predicates: buffer-live-p, pointer equality with eq, and buffer-local-value. Executing these checks 1000 times takes a fraction of a millisecond in Emacs Lisp. The heavy operations, such as diff-hl-remove-overlays and diff-hl–update-overlays, are gated behind (eq (diff-hl–target-buffer buf) base).

Right. Have you done any measurements?

```
(dotimes (i 10000)
  (save-excursion
    (switch-to-buffer (format " testbuf %s" i))))

(benchmark-run-compiled 1 (dolist (buf (buffer-list)) (diff-hl--target-buffer buf)))
```

shows me about 10ms of runtime (with occasional spikes to 65 ms with GC), which should average to 1ms per 1000 buffers. That's an extreme case, of course.

jamescherti:
Checking...

jamescherti:
My `(length (buffer-list))`: 10023. 

Here is my result for `(benchmark-run-compiled 1 (dolist (buf (buffer-list)) (diff-hl--target-buffer buf)))`:
```
(0.0018128579999999999 0 0.0)
```

In my setup, it takes under two milliseconds. The Lisp function call overhead is negligible.

Let’s use these results as the basis for future optimizations.

EDIT: All Emacs Lisp .el files are compiled natively.

dgutov:
Thanks, okay, let's leave it at that.

My measurement is higher, but it could be due to different cpu/optimizations/no-native-compile.

jamescherti:
Sounds good, Dmitry.

PR #278, opened 2026-05-20
Reduce flydiff-update calls in diff-hl-flydiff-mode

ermingol23:
When diff-hl-flydiff-mode is enabled, ensure that diff-hl-update function is only called when there are changes in the buffer and not every time the idle timer triggers. This reduces the number of calls to diff-hl-update and thus allows diff-hl-mode to run smoother on weaker (or older hardware).

The function diff-hl-flydiff/modified-p is rewritten so it uses buffer-chars-modified-tick internally instead of relying on buffer-modified-p. It is thus the only function that need to call buffer-chars-modified-tick and all other calls have been removed. Furthermore the buffer-local variable diff-hl-flydiff-modified-tick is only updated when it differs from buffer-chars-modified-tick.

Note that when diff-hl-flydiff/modified-p is called there is a guard that checks if the given state is one of 'added' or 'missing'. Unless one of those two states is given this will trigger a check using buffer-chars-modified-tick. This means that a nil state will trigger such a check.

The idle timer calls the diff-hl-update function, but an :around advice is used. That is the diff-hl-flydiff-update function decides if and when the original diff-hl-update function is called. To simplify the logic the existing 'unless' is replaced with 'when' and thus all checks are inverted; only when all conditions are met the update function diff-hl-update is called.

Customizeable variable diff-hl-flydiff-delay added to existing group diff-hl-flydiff.

Documentation strings written for existing functions diff-hl-flydiff-update and diff-hl-flydiff/modified-p.

NOTE: diff-hl-dired.el had a hard dependency to the vc-run-delayed-success function introduced in Emacs 31. A call to 'static-if' is added to check if such a function is bound or not. If it is not bound, then a direct call to vc-hg-after-dir-status is made instead. Not sure if this is the best way to support older Emacs versions.


diff-hl-flydiff.el
```diff
@@ -34,32 +34,40 @@
 
 (defcustom diff-hl-flydiff-delay 0.3
   "The idle delay in seconds before highlighting is updated."
-  :type 'number)
+  :type 'number
+  :group 'diff-hl-flydiff)
```
dgutov, line 38: `:group` is usually not needed if the last `defgroup` in the file is the same.

diff-hl-flydiff.el
```diff
@@ -34,32 +34,40 @@
...
-           (file-remote-p default-directory)
-           (not (file-exists-p buffer-file-name)))
-    (diff-hl-update)))
+(defun diff-hl-flydiff-update (orig-fun)
+  "Update diff markings if buffer has been modified since last call.
+
+Calls ORIG-FUN only when buffer has been modified since last
+call, buffer exists as a file, and it is not a remote file."
+  (when (and diff-hl-mode
+             (diff-hl-flydiff/modified-p nil)
```
dgutov, line 56: What happens if subsequent checks resolve to nil? But `diff-hl-flydiff/modified-p` already updated `diff-hl-flydiff-modified-tick`. Maybe we should do it here, before calling orig-fun.

dgutov (changes requested):
Things look great otherwise.

Could I ask if you are willing to sign FSF's copyright papers?

You submitted two PRs now, and while I could probably squeeze this one in, the two together are over the allowed limit for not assigned code.

ermingol23:
`Could I ask if you are willing to sign FSF's copyright papers?
`
Sure! No problems! Please let me know how to do it (what forms to fill in, where to send them, and so on).

diff-hl-flydiff.el
ermingol23: I'll remove the ':group'!

diff-hl-flydiff.el
ermingol23: Not entirely sure I follow you here. The construct looks like this

```
(when (and diff-hl-mode
             (diff-hl-flydiff/modified-p nil)
             (stringp buffer-file-name)
             (not (file-remote-p default-directory))
             (file-exists-p buffer-file-name))
    (apply orig-fun nil))
```
Since the 'and' function in Emacs LISP short-circuits, the following checks are never called if 'diff-hl-flydiff/modified-p' return 'nil'.

Are you thinking of a scenario where a buffer is created that is not associated with any file and then it is written to a file (function 'buffer-file-name' goes from returning 'nil' to a string containing name of file)? If you then insert/modify a character in the file 'diff-hl-flydiff/modified-p' would return non-'nil' and 'orig-fun' is called.

There might be a race condition though where a buffer is modified, the timer triggers a check, 'diff-hl-flydiff/modified-p' return non-'nil' and then the file is moved which causes the 'and'-clause to return 'nil' (due to that 'file-exists-p' would return 'nil'). If the file is moved back and the timer fires again the 'and'-clause would still return 'nil' due to that 'diff-hl-flydiff/modified-p' return 'nil' in this case.

One way of removing this race-condition would be to introduce an additional buffer-local variable (for instance 'diff-hl-flydiff--update-state') that is cleared after 'orig-fun' returns and 'diff-hl-flydiff/modified-p' do an 'or' between the value of this new variable and 'buffer-chars-modified-tick' like this

```
  (unless (memq state '(added missing))
    (let ((buffer-state (not (eq diff-hl-flydiff-modified-tick
                          (buffer-chars-modified-tick)))))
      (if (or buffer-state diff-hl-flydiff--update-state)
          (progn
            (setq diff-hl-flydiff-modified-tick (buffer-chars-modified-tick))
            (setq diff-hl-flydiff--update-state t)
            (setq buffer-state diff-hl-flydiff--update-state)))
      buffer-state)))
```

That is, if you time above attempt on causing a race condition as described the next time the timer fires 'diff-hl-flydiff-modified-tick' is still updated (even though the value has not changed) and non-'nil' is returned the next time the function is called. The following tick would then result in an update where 'orig-fun' is called and 'diff-hl-flydiff--update-state' is set to 'nil', that is

```
  (when (and diff-hl-mode
             (diff-hl-flydiff/modified-p nil)
             (stringp buffer-file-name)
             (not (file-remote-p default-directory))
             (file-exists-p buffer-file-name))
    (apply orig-fun nil)
    (setq diff-hl-flydiff--update-state nil))
```

Perhaps you were thinking of something else that I missed?

dgutov:
Great! Here you go:

```
Please email the following information to assign@gnu.org, and we
will send you the assignment form for your past and future changes.

Please use your full legal name (in ASCII characters) as the subject
line of the message.
----------------------------------------------------------------------
REQUEST: SEND FORM FOR PAST AND FUTURE CHANGES

[What is the name of the program or package you're contributing to?]
Emacs

[Did you copy any files or text written by someone else in these changes?
Even if that material is free software, we need to know about it.]


[Do you have an employer who might have a basis to claim to own
your changes?  Do you attend a school which might make such a claim?]


[For the copyright registration, what country are you a citizen of?]


[What year were you born?]


[Please write your email address here.]


[Please write your postal address here.]





[Which files have you changed so far, and which new files have you written
so far?]
```

diff-hl-flydiff.el
dgutov: Sorry about the wait. Let's take a step back: please describe a scenario in which the existing solution (that uses `diff-hl-flydiff-modified-tick`) does not satisfy our expectation.

Which scenario are we fixing? The simplest one I've just retested seem to do fine. You can add a message to the beginning of `diff-hl-update` to find the cases when it shouldn't fire but it does.

```diff
diff --git a/diff-hl.el b/diff-hl.el
index 15ba187..1aedf35 100644
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -640,6 +632,7 @@ contents as they are (or would be) after applying the changes in NEW."
 
 (defun diff-hl-update ()
   "Updates the diff-hl overlay."
+  (message "diff-hl-update called")
   (setq diff-hl-timer nil)
   (if (and (diff-hl--use-async-p)
            (eq diff-hl-update-async 'thread))
```

dgutov: In the meantime, I've pushed the two smaller fixes from this PR: cb049a2e8af7946669ef6b0c4e0cf6b188a54d33, 6b28f8ae77b5e60362c16d1e32a3b0baf41494cc

dgutov: And this takes care of the remaining one, I think: https://github.com/dgutov/diff-hl/commit/759e7f18bf8c92fdf742e678d360d660f6d41f10

dgutov:
I think for all three issues referenced here I either cherry-picked the fix or resolved another way.

@ermingol23 Please check out the latest version, and report if any issues remain.

diff-hl-flydiff.el
ermingol23: > Sorry about the wait. Let's take a step back: please describe a scenario in which the existing solution (that uses diff-hl-flydiff-modified-tick) does not satisfy our expectation.
> 
> Which scenario are we fixing? The simplest one I've just retested seem to do fine. You can add a message to the beginning of diff-hl-update to find the cases when it shouldn't fire but it does.
> 

Fair enough! The code I pushed originated from when I experienced severe performance problems running Emacs on older hardware (10+ years) and files were located on a slow NFS mount. What I attempt to do in the pull-request is to minimize the number of times flydiff calculates a diff as that was an expensive operation (preferably much fewer times than the timer fires).

Centralizing the handling of the 'diff-hl-flydiff-modified-tick' tick variable made it possible to have a single point where it was decided if the timer event should be acted on or not (the function 'diff-hl-flydiff/modified-p' seemed like a suitable spot for this and comparing against 'buffer-chars-modified-tick' is more efficient than calling 'buffer-modified-p' (I used Emacs profiling tool to conclude this when I made the implementation a couple of years ago).

If you feel my proposed code changes introduces more complexity than is warranted (or fixes something that is already fixed) I am OK with that. Not all fixes are good fixes "in the big picture", and it is good that you question the rationale behind a fix before blindly accepting it.  :-)

dgutov: > Centralizing the handling of the 'diff-hl-flydiff-modified-tick' tick variable made it possible to have a single point where it was decided if the timer event should be acted on or not

I think my last patch referenced above does that as well, but a little more compactly.

Could you try the same scenarios (old hardware, NFS mount) with the latest development version of diff-hl?

In either case, even if the code wasn't accepted as-is, having the issues reported is very valuable.

ermingol23: > I think my last patch referenced above does that as well, but a little more compactly.

Excellent!

> Could you try the same scenarios (old hardware, NFS mount) with the latest development version of diff-hl?

Alas I no longer have access to that old system. :-(

But I trust the issue I experienced is fixed now. :-)

PR #279, opened 2026-05-20
Add customizable inline hunk faces

ermingol23:
Add two cusomizable faces (four in total)

  - diff-hl-show-hunk-bg Background color for changed region when hunk is shown. This face is used for creating an overlay that tints the background for the current lines of the changed region in the buffer when viewing the hunk.

  - diff-hl-show-hunk-inline-base Foreground and background colors for the header and footer when viewing a hunk.

  - diff-hl-show-hunk-inline-ul Inherits from diff-hl-show-hunk-inline-base and adds an underline to the face.

  - diff-hl-show-hunk-inline-ol Inherits from diff-hl-show-hunk-inline-base and adds an overline to the face.

Modifies diff-hl-show-hunk--click function so mouse clicks toggles showing the hunk on and off.

### DIFF
--- a/diff-hl-show-hunk-inline.el
+++ b/diff-hl-show-hunk-inline.el
@@ -74,6 +74,24 @@ the hunk consist only on added lines, then
 `diff-hl-show-hunk--no-lines-removed-message' it is shown."
   :type 'boolean)
 
+(defface diff-hl-show-hunk-inline-base
+  '((((class color) (min-colors 88) (background light))
+     :foreground "black" :background "gray88" :extend t)
+    (((class color) (min-colors 88) (background dark))
+     :foreground "white" :background "#gray13" :extend t))
+  "Face used for background of shown hunk."
+  :group 'diff-hl-show-hunk-inline)
+
+(defface diff-hl-show-hunk-inline-ul
+  '((((class color) (min-colors 88))
+     :inherit diff-hl-show-hunk-inline-base :underline t))
+  "Face used for background of shown hunk with underline.")
+
+(defface diff-hl-show-hunk-inline-ol
+  '((((class color) (min-colors 88))
+     :inherit diff-hl-show-hunk-inline-base :overline t))
+  "Face used for background of shown hunk with overline.")
+
 (defun diff-hl-show-hunk-inline--splice (list offset length)
   "Compute a sublist of LIST starting at OFFSET, of LENGTH."
   (butlast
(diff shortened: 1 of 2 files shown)


dgutov:
Hey, this is interesting.

Some notes:

* `diff-hl-show-hunk-inline-base` should either have "white" as the fg for dark themes, or probably no `:foreground` specification at all.
* Its background looks like the active mode-line to me. Should it inherit from `mode-line-active`?
* `old lace` is not ideal, but it's workable. Hope people take a look and suggest more options.
* Other things being equal it's better to use named colors, I think.
* Please don't add consecutive empty lines

dgutov:
If anybody else looks at this PR, a picture:

image

ermingol23:
> - diff-hl-show-hunk-inline-base should either have "white" as the fg for dark themes, or probably no :foreground specification at all.

OK!

> - Its background looks like the active mode-line to me. Should it inherit from mode-line-active?

I did not realize this before you mentioned it (that the grey color I settled on actually matches closely the default active mode-line). I tried selecting a neutral color that would not stand out but still worked with the "standard" red/green diff-colors and at the same time separated from the rest of the buffer. With that said I am not sure it would be such a good idea to connect these (grey) separators with the color of the active mode-line. What happens if the active mode-line is set to a color that is close to one of the diff-colors, the 'diff-hl-show-hunk-face', or clashes severely with the other colors? My gut-feeling tells me that it is a good idea to have full control over the selected colors within diff-hl for things like separators as you could otherwise get "interesting" surprises when relying on other already existing face definitions.

> - old lace is not ideal, but it's workable. Hope people take a look and suggest more options.

I admit that I struggled with this one. I finally settled on it in an attempt for a neutral off-white color for the region representing what is within the buffer so it stands out against the diff (the idea was to highlight what the diff is about) but at the same time have a low risk of interfering with any font-locking going on. I agree wholeheartedly on that 'old lace' is not ideal, but I had a hard time coming up with something better... I am of course open to other suggestions here as well. :-)

> - Other things being equal it's better to use named colors, I think.

Ok!

> - Please don't add consecutive empty lines

Sorry about that! Will remove in the updated PR.

dgutov:
> What happens if the active mode-line is set to a color that is close to one of the diff-colors, the 'diff-hl-show-hunk-face', or clashes severely with the other colors?

I suspect that it won't: any consistent theme will have separate diff mode colors and the mode-line.

We don't have to adopt mode line's face specifically. Just if we don't, that creates a case of so-called "uncanney valley" with the colors beeing too close but not the same, and without paricular artistic purpose in that difference.

We can try some other built-in face with related gamma.

> I admit that I struggled with this one. I finally settled on it in an attempt for a neutral off-white color for the region representing what is within the buffer so it stands out against the diff (the idea was to highlight what the diff is about) but at the same time have a low risk of interfering with any font-locking going on. I agree wholeheartedly on that 'old lace' is not ideal, but I had a hard time coming up with something better... I am of course open to other suggestions here as well. :-)

Right. I think we'll merge this face as-is, and whatever feedback arrives will be after.

PR #280, opened 2026-05-26
Fix: diff-hl-magit-post-refresh wrong file name

danielfleischer:
`git diff-tree` is used and generates relative (to top level) file names. 
`buffer-file-name` generates full paths. 

The following condition always fails due to the file names being different:

https://github.com/dgutov/diff-hl/blob/7d873b2f58908de1ea2f499da9bf993e088953d7/diff-hl.el#L1440

This fixes the refresh related to magit actions.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -1437,7 +1437,7 @@ The value of this variable is a mode line template as in
                    (backend (vc-backend file)))
               (when backend
                 (cond
-                 ((member (file-name-nondirectory file) modified-files)
+                 ((member (file-relative-name file topdir) modified-files)
                   (when (memq (vc-state file) unmodified-states)
                     (vc-state-refresh file backend))
                   (diff-hl-update))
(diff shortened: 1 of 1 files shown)


dgutov:
Hi!

`(file-name-nondirectory file)` will return the file name stripped of all directory parts.

Maybe you wanted to use `file-relative-name`?

I suggest you also add a message in there printing the values on mismatch and post it the result here (with the fix not applied yet). To document the failure as it happens.

danielfleischer:
Thank you.

First, you are correct, I focused on top-level when testing. Fixed in 581c577.

Second, when debugging the cond branch with:

```elisp
(or
 (member file modified-files)
 (ignore (message "FILE %s NOT IN %s" file modified-files)))
```

one gets:

```
FILE /Users/dan/Documents/Projects/my_project/fileA NOT IN (fileA)
```

hence the filename normalization.

PR #283, opened 2026-06-05
Optimize diff-hl--update

jamescherti:
This pull request optimizes `diff-hl--update`.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -744,20 +744,21 @@ Return a list of line overlays used."
             (let ((ref-changes (diff-hl-adjust-changes ref-changes changes))
                   (base (diff-hl--target-buffer orig)))
               (dolist (buf (buffer-list))
-                (when (and (buffer-live-p buf)
-                           (eq (diff-hl--target-buffer buf) base)
+                (when (and (eq (diff-hl--target-buffer buf) base)
                            (buffer-local-value 'diff-hl-mode buf))
                   (with-current-buffer buf
+                    (diff-hl-remove-overlays)
                     (let (reuse)
-                      (diff-hl-remove-overlays)
-                      (let ((diff-hl-highlight-function
-                             diff-hl-highlight-reference-function)
-                            (diff-hl-fringe-face-function
-                             diff-hl-fringe-reference-face-function))
-                        (setq reuse (diff-hl--update-overlays ref-changes nil)))
-                      (diff-hl--update-overlays changes reuse)
-                      (when (not (or changes ref-changes))
-                        (diff-hl--autohide-margin))))))))))))))
+                      (when ref-changes
+                        (let ((diff-hl-highlight-function
+                               diff-hl-highlight-reference-function)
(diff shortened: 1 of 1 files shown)


jamescherti:
This adds two optimizations during the buffer traversal.

First, `(eq buf base)` is checked before calling `(buffer-base-buffer buf)`. This short-circuits evaluation to bypass the `buffer-base-buffer` call overhead.

Second, the boolean state of `(or changes ref-changes)` outside the loop. This prevents the Lisp interpreter from recalculating the same static condition for all active buffers.

diff-hl.el
```diff
@@ -741,22 +741,26 @@ Return a list of line overlays used."
         (lambda (ref-changes)
           (when (buffer-live-p orig)
             (let ((ref-changes (diff-hl-adjust-changes ref-changes changes))
-                  (base (diff-hl--target-buffer orig)))
+                  (base (diff-hl--target-buffer orig))
+                  (has-changes (or changes ref-changes)))
               (dolist (buf (buffer-list))
                 (when (and (buffer-live-p buf)
-                           (eq (diff-hl--target-buffer buf) base)
+                           (or (eq buf base)
```
dgutov: Moving `(eq buf base)` up is solid

dgutov:
Hi!

I like the `(when ref-changes` and `(when changes` improvements.

The other two seem negligible: considering we save the `buffer-base-buffer` call only once - when the buffer matches the original, but not for the other buffers; and we break the "target buffer" abstraction. Saving an `or` when the buffer already matched won't move the needle.

We could also drop the `(buffer-live-p buf)` check inside the loop because buffer-list returns only live ones. And that probably isn't going to change during our execution - we don't allow to run any timers or process filters during `diff-hl--update-overlays` or `diff-hl--autohide-margin`.

jamescherti:
That makes perfect sense. I have updated the pull request to preserve the diff-hl--target-buffer abstraction and dropped the redundant buffer-live-p checks inside the loop.

jamescherti:
I also set inhibit-redisplay to t during the resolution callback to prevent the display engine from rendering intermediate UI states during asynchronous overlay updates. Because the function recreates a large number of overlays across multiple buffers, the Emacs display engine can waste CPU cycles recalculating display properties, margins, or fringe bitmaps before the buffer state is completely finalized.

dgutov:
> I have updated the pull request to preserve the diff-hl--target-buffer abstraction and dropped the redundant buffer-live-p checks inside the loop.

Please drop the `or` optimization too, as I mentioned in the previous message.

Regarding `inhibit-redisplay` - do you have any measurements that show an improvement? It does help in some scenarios, but here I expect no redisplays to happen during that block's execution.

Redisplay happens at specific points of the event loop, not right after any change in a buffer.

jamescherti:
I have removed the 'or' and `inhibit-redisplay`.

dgutov:
Thanks!

jamescherti:
You're very welcome!

Issue #284, opened 2026-06-05
Capture buffer not associated with a file


ak-ionos:
Hello,

the latest diff-hl changes lead to the following error when attempting to org-capture an entry using file+olp+datetree. I checked out rev. 902d082d70b962103342ee46e73fc6886bc8720a, which still works fine, so I suspect the indirect buffer changes to be the culprit.

```
Debugger entered--Lisp error: (error "Buffer CAPTURE-sports.org is not associated with a file")
  error("Buffer %s is not associated with a file" "CAPTURE-sports.org")
  vc-ensure-vc-buffer()
  diff-hl-diff-buffer-with-reference("/home/ak/src/org/sports.org" " *diff-hl-diff*" Git)
  diff-hl-flydiff-changes-buffer("/home/ak/src/org/sports.org" Git nil nil)
  apply(diff-hl-flydiff-changes-buffer ("/home/ak/src/org/sports.org" Git))
  diff-hl-changes-buffer("/home/ak/src/org/sports.org" Git)
  diff-hl-changes()
  diff-hl--update()
  diff-hl-update()
  diff-hl-flydiff-update()
  apply(diff-hl-flydiff-update nil)
  timer-event-handler([t 0 0 299999 t diff-hl-flydiff-update nil idle 999999 nil])
```

I run global-diff-hl-mode upon init, but never had to exclude CAPTURE buffers before.

dgutov:
Hi! Thanks for the report.

@jamescherti Please take a look.

jamescherti:
@ak-ionos @dgutov Could you please try this pull request https://github.com/dgutov/diff-hl/pull/285 and confirm that it fixes the issue?

dgutov:
Could it actually be a case when we don't want the diff-hl highlights to be shown?

I.e. indirect buffer is used for narrowed editing, and that's it.

dgutov:
What would be the ideal behavior? @ak-ionos

jamescherti:
IMHO, the highlights should be displayed in narrowed indirect buffers (e.g., `org-capture`).

And the pull request https://github.com/dgutov/diff-hl/pull/285 fixes this issue for me:

<img width="2090" height="309" alt="Image" src="https://github.com/user-attachments/assets/ef1bf77a-c8eb-494b-b83a-ad28d436e758" />

ak-ionos:
Using diff-hl from your branch fixes the bug for me. Thanks a lot, @jamescherti !

[...]

PR #285, opened 2026-06-05
Fix VC function compatibility in indirect buffers

jamescherti:
Some VC functions (e.g., vc-buffer-sync) rely on the current buffer's file association and local state. When invoked from an indirect or cloned buffer, these functions can fail or return incorrect results.

This pull request wraps these buffer-dependent VC calls in (with-current-buffer (diff-hl--target-buffer ...)) to ensure they always execute in the context of the base buffer.

The diff-hl--target-buffer function was also updated to make its argument optional, cleaning up several redundant calls.

Bump emacs Package-Requires to "27.1" to guarantee indirect buffer support in VC functions.

Add a safety check to diff-hl-diff-buffer-with-reference, which ensures the buffer is visiting a file before interacting with version control

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -6,7 +6,7 @@
 ;; URL:      https://github.com/dgutov/diff-hl
 ;; Keywords: vc, diff
 ;; Version:  1.10.0
-;; Package-Requires: ((cl-lib "0.2") (emacs "26.1"))
+;; Package-Requires: ((cl-lib "0.2") (emacs "27.1"))
 
 ;; This file is part of GNU Emacs.
 
@@ -314,14 +314,15 @@ It can be a relative expression as well, such as \"HEAD^\" with Git, or
      (lambda (value)
        (or (null value) (stringp value))))
 
-(defun diff-hl--target-buffer (buf)
+(defun diff-hl--target-buffer (&optional buf)
   "Return the correct buffer for the situation, preferring the base buffer."
-  (or (buffer-base-buffer buf) buf))
+  (let ((buf (or buf (current-buffer))))
+    (or (buffer-base-buffer buf) buf)))
 
 (defun diff-hl--buffer-file-name (&optional buffer)
   "Return the file name of the BUFFER or its base buffer.
 BUFFER defaults to the current buffer."
-  (buffer-file-name (diff-hl--target-buffer (or buffer (current-buffer)))))
+  (buffer-file-name (diff-hl--target-buffer buffer)))
 
 (defun diff-hl-define-bitmaps ()
   (let* ((scale (if (and (boundp 'text-scale-mode-amount)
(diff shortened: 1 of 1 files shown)


jamescherti:
I tested this pull request with org-capture and it works:
image

jamescherti:
I have also made changes to fix both `diff-hl-revert-hunk-1` and `diff-hl-after-checkin`.

jamescherti:
This pull request fixes the issue: https://github.com/dgutov/diff-hl/issues/284#issuecomment-4634261654

dgutov:
@jamescherti Remember the several places we have the redundant `(with-current-buffer (current-buffer` now?

How about we try to localize the fix there?

Both `diff-hl-revert-hunk` and `diff-hl-diff-goto-hunk` are in that set, for instance.

jamescherti:
Hello @dgutov,

Let's merge this fix and then work on enhancements in a separate pull request.

Merging this will fix https://github.com/dgutov/diff-hl/issues/284 for the users who are experiencing it. It also fixes the issue with reverting the hunk.

dgutov:
@jamescherti Could we limit the fix to those `with-current-buffer` forms? Or do we have to add new ones?

How do we reproduce the issue? Just using `M-x org-capture` in a version-controlled .org file?

dgutov:
~If we have to add a new one, we can add it in `diff-hl-flydiff-update`.~

~Or change `save-current-buffer` to `with-current-buffer (...)` in `diff-hl-diff-buffer-with-reference`.~

~But preferably just one of these.~

dgutov:
(Never mind the above.)

I've pushed a small fix for the linked issue, so let's consider these changes separately.

Some comments:

* `vc-backend` shouldn't need a change of buffer because it gets a file name already.
* `vc-buffer-sync-fileset` - probably same.
* `vc-deduce-fileset` uses the base buffer since 2020 - should be fine without the change (raising dependency to Emacs 27 or 28 is fine).

jamescherti:
FYI: I've resolved the conflict and merged your fix (`vc-backend` in `diff-hl-diff-buffer-with-reference`) into this pull request.

jamescherti:
I removed the redundant `with-current-buffer` wrappers around `vc-buffer-sync-fileset`, `vc-deduce-fileset`, `vc-deduce-backend`, and `vc-backend`, I bumped `emacs` Package-Requires to "27.1", and I added a safety check to `diff-hl-diff-buffer-with-reference`, which ensures the buffer is visiting a file before interacting with version control functions, replacing the removed `vc-ensure-vc-buffer`.

dgutov:
Check the "merged with master" status, preferably rebase.

And it seems we need a new name for the PR and the commit description. Fixing two commands in indirect buffers?

jamescherti:
I resolved the conflict.

dgutov:
Thanks!

jamescherti:
My pleasure, @dgutov!

PR #290, opened 2026-06-21
diff-hl-show-hunk-inline: Put underline at descent line

emreyolcu:
Currently the inline popups display the top border (drawn using `:underline` on the header line) slightly too high. This is because Emacs draws underlines at the baseline unless `x-underline-at-descent-line` is enabled, but that setting affects Emacs globally. Since Emacs 29, the `:underline` attribute accepts the property `:position`, which specifies where the underline should be drawn. If it is non-nil, then the underline is displayed at the descent of the text instead of the baseline. I've made a tiny change to enable that property for underlines in diff-hl when Emacs 29+ is detected. I've included two screenshots below showing the difference. Notice that the bottom border is unaffected, because it's an overline.

**Before:**

Screenshot From 2026-06-21 16-02-00

**After:**

Screenshot From 2026-06-21 16-01-11

### DIFF
--- a/diff-hl-show-hunk-inline.el
+++ b/diff-hl-show-hunk-inline.el
@@ -105,6 +105,10 @@ Compute it from LINES starting at INDEX with a WINDOW-SIZE."
          (index (min index (- len window-size))))
     (diff-hl-show-hunk-inline--splice lines index window-size)))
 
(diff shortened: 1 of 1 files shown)


dgutov:
Looks good, thanks!

PR #291, opened 2026-06-21
diff-hl-show-hunk-inline: Enable customization of scroll indicators

emreyolcu:
I've made the scroll indicators shown for inline popups customizable (the default choices look bad with certain fonts in my opinion). Let me know if you'd like to merge this but want it done differently (spaces could be made part of the customizable indicators for instance).


dgutov:
Hi, thank you.

Including the spaces might be a good idea - the user then could use a wider char, or a combination of such, for the arrows.

emreyolcu:
Rebased onto master and updated to let indicators be strings instead of chars. What do you think about this version?

dgutov:
Looks good. Thanks!

PR #293, opened 2026-07-28
Use jj `commit_id`, not `change_id`, as the reference revision

jaybosamiya:
`diff-hl-create-revision` caches the reference file in a temp file keyed by the revision string, assuming that string is content-addressed, and only refetches when it's missing. A jj change ID survives changes to the contents. So after (for example) squashing into `@-`, the fringe keeps showing the squashed lines as uncommitted. `commit_id` changes with the content and fixes it.

Related: 6938d20 / #250


dgutov:
Thanks!

PR #297, opened 2026-08-23
Handle unborn Git branches when staged changes are hidden.

### LINKED ISSUE #298: Error when hiding staged changes in a Git repository without commits
## Description

diff-hl fails when it hides staged changes in a Git repository that has
no commits.

The failure occurs when all of these conditions are true:

- The repository has no `HEAD`.
- A file has staged and unstaged changes.
- `diff-hl-show-staged-changes` is nil.
- `diff-hl-highlight-reference-function` is non-nil.

## Steps to reproduce

Create a Git repository with one staged change and one unstaged change:

```sh
git init
printf 'staged\n' > tracked
git add tracked
printf 'staged\nworking\n' > tracked
```

Visit `tracked` in Emacs and evaluate:

```elisp
(setq diff-hl-show-staged-changes nil
      diff-hl-update-async t)

(diff-hl-mode 1)
(diff-hl-update)
```

## Actual behavior

An asynchronous update reports this error:

```text
diff-beginning-of-hunk: Can’t find the beginning of the hunk
```

A synchronous update exposes the failed Git command:

```elisp
(let ((diff-hl-update-async nil))
  (diff-hl-update))
```

```text
Failed (status 128): git --no-pager diff-index -U0 -p --cached HEAD -- tracked
```

diff-hl runs `git diff-index --cached HEAD` to calculate the staged
changes. The command fails because the repository has no `HEAD`.

During an asynchronous update, diff-hl passes the Git error output to
`diff-hl-changes-from-buffer`. This function expects a unified diff. It
fails because the output does not contain a diff hunk header.

## Expected behavior

diff-hl should show the staged and unstaged changes without reporting an
error.

Git supports repositories without commits through `git diff --cached`
when no commit is specified. This command uses `HEAD` when it exists. It
uses the empty tree when `HEAD` does not exist.

## Environment

- Emacs 30.2
- diff-hl revision `6f1df2b83d1140a2938409b35de7f4c9c7b1defd`
- `diff-hl-highlight-reference-function` is non-nil

am5083:
Fixes #298 

When `diff-hl-show-staged-changes` is nil, diff-hl currently
computes the staged/reference layer with:

    git diff-index --cached HEAD

This fails in a repository with an unborn branch because `HEAD`
does not exist yet. Synchronous updates signal status 128.
Asynchronous updates subsequently pass Git's fatal diagnostic to
`diff-hl-changes-from-buffer`, which fails in
`diff-beginning-of-hunk`.

Use `git diff --cached [<commit>]` instead. When no explicit commit
is supplied, Git compares the index against `HEAD` when it exists
and against the empty tree on an unborn branch. An explicit
`diff-hl-reference-revision` is still passed through unchanged.

The regression test creates a temporary unborn repository with
separate staged and worktree changes and verifies both layers with
synchronous and asynchronous updates. The existing test also
checks implicit and explicit `HEAD` references.

Tested with Emacs 30.2:

- staged-change tests: 3/3 passed
- remaining applicable ERT tests: 15/15 passed
- byte compilation with warnings as errors passed
- `git diff --check` passed

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -476,16 +476,18 @@ BUFFER defaults to the current buffer."
            "diff-files"
            (cons "-p" (vc-switches 'git 'diff))))
    ((eq new-rev 'git-index)
+    ;; With no revision, `git diff --cached' uses HEAD when it exists and
+    ;; the empty tree on an unborn branch.
     (apply #'vc-git-command buffer
            (if (diff-hl--use-async-p) 'async 1)
            (list file)
-           "diff-index"
+           "diff"
            (append
             (vc-switches 'git 'diff)
-            (list "-p" "--cached"
-                  (or diff-hl-reference-revision
-                      (diff-hl-head-revision backend))
-                  "--"))))
+            (list "-p" "--cached")
+            (when diff-hl-reference-revision
+              (list diff-hl-reference-revision))
+            (list "--"))))
    (t
     (condition-case err
         (vc-call-backend backend 'diff (list file)
--- a/test/diff-hl-test.el
+++ b/test/diff-hl-test.el
@@ -168,12 +168,14 @@
       (should
        (null
         (assoc-default :reference (diff-hl-changes)))))
-    (let* ((diff-hl-show-staged-changes nil)
-           (res-buf (assoc-default :reference (diff-hl-changes))))
-      (should
-       (equal
-        (diff-hl-changes-from-buffer res-buf)
-        '((1 1 0 insert)))))))
+    (dolist (reference '(nil "HEAD"))
+      (let* ((diff-hl-reference-revision reference)
+             (diff-hl-show-staged-changes nil)
+             (res-buf (assoc-default :reference (diff-hl-changes))))
+        (should
+         (equal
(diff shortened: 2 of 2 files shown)


diff-hl.el
```diff
@@ -476,16 +476,18 @@ BUFFER defaults to the current buffer."
...
+    ;; the empty tree on an unborn branch.
     (apply #'vc-git-command buffer
            (if (diff-hl--use-async-p) 'async 1)
            (list file)
-           "diff-index"
+           "diff"
            (append
             (vc-switches 'git 'diff)
-            (list "-p" "--cached"
-                  (or diff-hl-reference-revision
```
dgutov, line 486: Hi! Could you try this version?

```el
(or diff-hl-reference-revision
      (vc-git--rev-parse "HEAD")
      "4b825dc642cb6eb9a060e54bf8d69288fbee4904")
```

still with `diff-index`, without switching to `git diff`. The above literal is the "empty repository" hash for SHA-1, we use it below too.

dgutov: Hmm, or maybe not - that adds +1 process call on the common path

PR #300, opened 2026-09-04
Make diff-hl-timer buffer-local

NicholasBHubbard:
This PR makes `diff-hl-timer` buffer-local so a pending timer in one buffer does not prevent `diff-hl-update-once` from scheduling an update in another.

I discovered this while using `(setq diff-hl-update-async t)` and I noticed that newly opened buffers would not receive highlights until their first save.

I added a test that verifies that `diff-hl-update-once` schedules an independent timer for each buffer.

### DIFF
--- a/diff-hl.el
+++ b/diff-hl.el
@@ -638,7 +638,7 @@ contents as they are (or would be) after applying the changes in NEW."
         (run-hook-with-args-until-success 'diff-hl-async-inhibit-functions
                                           default-directory))))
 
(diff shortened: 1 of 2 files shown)


dgutov:
Hi! Thanks for this. 

I've added a small follow-up which ensures that the update is only done once. Otherwise each applicable major mode in derived-mode chain invokes it (debouncing this is the reason the timer was introduced in the first place).

